PR #7, opened 2015-02-12
Support --profile-disable-exp

mgorny:
Alike -dev. Travis kinda fails but it seems to be some other issue. Tests pass locally for me.


radhermit:
Yes, that git log parsing for version stuff in snakeoil needs to be tossed at some point.

PR #14, opened 2015-06-15
LicenseAddon: support finding licenses in parent repos

mgorny:
Make LicenseAddon traverse parent repositories looking for license directories. Also remove the meaningless --license-dir option.

### DIFF
--- a/pkgcheck/addons.py
+++ b/pkgcheck/addons.py
@@ -406,35 +406,29 @@ class StableCheckAddon(base.Template):
 class LicenseAddon(base.Addon):
 
     @staticmethod
-    def mangle_option_parser(parser):
-        parser.add_option(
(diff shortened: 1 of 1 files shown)


radhermit:
Thanks, merged with some minor simplifications. I'll look into moving this support into pkgcore properly later on.

radhermit:
Actually this breaks tests so it'll have to be reworked a bit sooner rather than later.

radhermit:
Should be fixed via https://github.com/pkgcore/pkgcore/commit/e285feb90ab0ad4e33931ab4367efde1626ee349 and https://github.com/pkgcore/pkgcheck/commit/bc0f6818d04fe861b12d9b2a6cfb7e995fa30501.

PR #16, opened 2015-08-13
test: change FakePkg.local_use to dict to follow pkgcore

mgorny:
Following pkgcore/pkgcore#117.


radhermit:
Done slightly differently since an immutable dict is used instead.

PR #22, opened 2015-12-02
XmlReporter: include report class name

mgorny:
You won't like it but I guess there's no point in making this class any prettier now that you're working on a better thing. But until then, it'd be nice to have class name to make the output at least a bit machine-useful.


radhermit:
Merged.

PR #29, opened 2016-04-17
XML Schema for metadata.xml validation

mgorny:
Since DTD is not powerful enough to perform real validation of metadata.xml, here's an update to use the official schemas instead.


mgorny:
And now with proper validator error message passing. However, I've implemented it only for libxml2. I'd appreciate if you could handle `spawn()` for xmllint yourself ;-).

mgorny:
Let's focus on #30 instead as it is cleaner.

PR #34, opened 2016-05-27
UseAddon: use profile-derived list of implicit USE flags

mgorny:
Use profile-derived list of implicit USE flags (that follows either EAPI
5 or hacky logic for earlier EAPIs) rather than the pre-EAPI 5 hardcoded
logic used so far.

This improves unused global USE flag check to include unused USE_EXPAND
flags. It also triggers on supposedly-implicit USE_EXPAND flags that are
not in USE_EXPAND_VALUES_*.

It may be a good idea to split the report in the future, to include
separately: unused global USE flags, unused (whole) USE_EXPANDs, unused
USE_EXPAND flags and missing USE_EXPAND_VALUES_*.

-------

Note: global CI run not done yet. I'm curious as to result due to the mint profile-induced breakage.

### DIFF
--- a/pkgcheck/addons.py
+++ b/pkgcheck/addons.py
@@ -462,26 +462,37 @@ class UnstatedIUSE(base.Error):
 
 class UseAddon(base.Addon):
 
+    required_addons = (ProfileAddon,)
     known_results = (UnstatedIUSE,)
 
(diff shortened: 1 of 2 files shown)


radhermit:
Merged with test fixes and minor changes.

PR #38, opened 2016-06-17
metadata_checks: Checks for := use in || () blocks and blockers

mgorny:


mgorny:
Ping? :-P

radhermit:
Thanks, merged with a couple changes: collapsing multiple errors into one report and adding some initial test cases.

PR #39, opened 2016-06-24
Fixes to Manifest checks

mgorny:
Fix a few solid bugs in Manifest checks that prevented conflicting checksum check from being run, then fix the check to not explode when trying to report.


mgorny:
Ping.

radhermit:
Merged... looks like I should schedule writing some test code for repo metadata checks. :smile:

PR #42, opened 2016-07-01
RestrictedRepoSource: report MetadataExceptions early and explicitly

mgorny:
Filter packages failing with MetadataException out
in RestrictedRepoSource, and report explicit MetadataLoadErrors for
them. This way the specific checks do not have to account for possible
MetadataExceptions and the problems are reported in check results rather
than only on console.


mgorny:
Ping.

radhermit:
I'd like to handle this better within pkgcore itself to improve other tooling as well instead of handling it separately everywhere.

PR #44, opened 2016-07-28
remove BlockMemoryExhaustion exception

4eetah:
pkgcore transitive use atom bug fixed


den4ix:
This should be merged after finishing with pkgcore#80

PR #55, opened 2017-06-15
Add a check for empty files in FILESDIR

mgorny:


radhermit:
I'd prefer if this came with at least a simple unit test.

mgorny:
I don't think there's any infra for text fixtures right now, and the check accesses ebuild directories directly. I would prefer if you didn't dump inventing some way of doing this properly on me. I wouldn't mind adding a test if there was at least one test for pkgdir-related checks in place already.

mgorny:
Ok, figured out how to add a test and added it. Hope you'll like the extra classes ;-P.

radhermit:
Sorry for the delay, thanks for writing the checks they're good enough to start with.

Merged with minor changes.

PR #58, opened 2017-06-19
metadata_checks: Check for complex nested REQUIRED_USE

mgorny:
Check for overcomplex nested REQUIRED_USE constructs according to
the preliminary GLEP for automated REQUIRED_USE solving. The reported
constructs include unnecessary all-of groups and ||/^^/?? groups that do
not contain flat flags.


radhermit:
As with the other new check, I'm generally trying to add new checks with at least a minimal test of some kind. I can/could do them for these, but it's not a high priority for me atm, i.e. if you want things merged faster some unit tests would be nice. :smile:

mgorny:
Considering this one stalled anyway since I'm waiting for GLEP no assignment, and I'd like to include the GLEP no in the check.

PR #65, opened 2017-12-17
Add check for deprecated checksums

mgorny:
Since MissingChksum no longer checks for generic checksum set mismatch
but only for missing required checksums, add a new warning-level check
that finds files whose checksums use old set.


radhermit:
Thanks, merged in https://github.com/pkgcore/pkgcheck/commit/b697b04c6b9cd1399273c1cbf9d3d52a95777f4d with some minor changes/fixes.

PR #67, opened 2018-01-06
Separate NonSolvableDeps classes for different profile types

mgorny:


mgorny:
Rebased for the modern code. @radhermit, could you take a look at it?

radhermit:
I'd rather rework profile objects in pkgcore a bit to add status as an attribute instead of having to create a new mapping.

radhermit:
Thanks, merged with some changes because the pkgcore rework appears to be taking too long to do what I wanted. It shouldn't be hard to update this whenever I finish the rework.

PR #70, opened 2018-04-14
metadata_xml: Do not report empty for <stabilize-allarches/>

mgorny:
The <stabilize-allarches/> element is empty by design, so it's not
an error for it to be empty ;-).


radhermit:
Thanks, merged in c399d95c2661806f0e134a24d3342cc100d5aa10.

PR #84, opened 2019-07-20
python-r1 checks

mgorny:
Live pull request to include tests specific to Python ecosystem in Gentoo.

### DIFF
--- a/src/pkgcheck/checks/python.py
+++ b/src/pkgcheck/checks/python.py
@@ -235,3 +235,12 @@ class PythonReport(base.Template):
                     if p.key in INTERPRETERS:
                         yield PythonRuntimeDepInAnyR1(pkg, attr, p)
                         break
+            for attr in ("depend", "bdepend"):
+                for p in iflatten_instance(getattr(pkg, attr), atom):
+                    if p.key in INTERPRETERS:
+                        break
+                else:
+                    continue
+                break
+            else:
+                yield PythonMissingDeps(pkg, 'DEPEND')
--- a/src/pkgcheck/checks/python.py
+++ b/src/pkgcheck/checks/python.py
@@ -100,6 +100,26 @@ class PythonMissingDeps(base.Warning):
                 f"in {self.dep_type}")
 
 
+class PythonRuntimeDepInAnyR1(base.Warning):
+    """Package depends on Python at runtime but uses any-r1 eclass."""
+
+    __slots__ = ("category", "package", "version", "dep_type", "dep")
+
+    threshold = base.versioned_feed
+
+    def __init__(self, pkg, dep_type, dep):
+        super().__init__()
+        self._store_cpv(pkg)
+        self.dep_type = dep_type
+        self.dep = dep
+
+    @property
+    def short_desc(self):
+        return (f"Package inherits python-any-r1 but has {self.dep_type} "
+                f"on {self.dep}; python-r1 or python-single-r1 should "
+                f"be used instead")
+
+
 class PythonReport(base.Template):
     """Python eclass issue scans.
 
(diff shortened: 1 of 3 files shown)


radhermit:
I think some of these are throwing false positives at the moment. For example, csound has PythonMissingRequiredUSE reported but it looks fine to me. It's just conditionalized behind the python USE flag.

src/pkgcheck/checks/python.py
```diff
@@ -40,6 +43,26 @@ def short_desc(self):
...
+    def __init__(self, pkg, flags, single_flags):
+        super().__init__()
+        self._store_cpv(pkg)
+        self.flags = flags
+        self.single_flags = single_flags
+
+    @property
+    def short_desc(self):
+        return (f"Python package has mismatched Python flags in IUSE: "
+                f"PYTHON_TARGETS={sorted(self.flags)} but "
```
radhermit: The flag sorting should be done during initialization and make them tuples since all result attributes should be immutable.

tests/checks/test_python.py
```diff
@@ -239,6 +277,139 @@ def test_missing_required_use(self):
...
                                  'python_single_target_python3_6',
+                            RDEPEND='python_single_target_python2_7? ( '
+                                    '  dev-lang/python:2.7 ) '
+                                    'python_single_target_python3_6? ( '
+                                    '  dev-lang/python:3.6 )',
                             REQUIRED_USE='|| ( python_targets_python2_7 '
                                          '  python_targets_python3_6 )')),
             python.PythonMissingRequiredUSE)
+
+    def test_missing_deps(self):
```
radhermit: Choosing this as a random place to comment on all the tests, it would be nice to add some additional comments between the various assertions (or groups of similar assertions) that describe what makes them different then the other ones under the same method if applicable.

Alternatively, you could also add error messages to be output on test failure where it makes sense.

src/pkgcheck/checks/python.py
```diff
@@ -80,6 +100,36 @@ def get_python_eclass(pkg):
         assert(len(eclasses) <= 1)
         return eclasses.pop() if eclasses else None
 
+    def scan_tree_recursively(self, deptree, expected_cls):
```
radhermit: As we discussed on briefly on IRC, I think functionality for this should be added to relevant classes in pkgcore. I can possibly look into that this coming week.

mgorny:
I'm looking into the results one by one. I'll try to figure out what makes csound unhappy.

mgorny:
Ok, found the mistake in the logic. When || was collapsed into parent block, it didn't properly skip non-python flags from that block.

tests/checks/test_python.py
mgorny: Added some comments where tests were non-obvious. Figured it doesn't make sense to say that test for missing PYTHON_DEPS with entirely missing RDEPEND needs special description ;-). Please let me know if you find any of the remaining tests confusing.

radhermit:
Merged with a bunch of changes on top. I fixed one false positive and found two more types:

dev-db/mongodb: shouldn't trigger PythonMissingDeps for RDEPEND since it's just using python for build support afaics
sys-kernel/aufs-sources: probably shouldn't trigger PythonMissingDeps for DEPEND but the python dep is buried in the kernel-2 eclass under an unset conditional for regular source-based builds.

I could start opening issues but those are just the ones I noticed on a quick run over the tree with the new check.

mgorny:
> dev-db/mongodb: shouldn't trigger PythonMissingDeps for RDEPEND since it's just using python for build support afaics

It's not a false positive, it's wrong eclass used.

> sys-kernel/aufs-sources: probably shouldn't trigger PythonMissingDeps for DEPEND but the python dep is buried in the kernel-2 eclass under an unset conditional for regular source-based builds.

kernel-2.eclass is an impossible mess that's probably not worth addressing right now. They could probably inherit python-any-r1 conditionally to their magical variable.

radhermit:
On another glance, the aufs-sources one appears correct and the eclass should be changed to only inherit the related python eclass when it actually needs those deps.

PR #92, opened 2019-07-25
Add check for obsolete GitHub/GitLab URIs

mgorny:


src/pkgcheck/checks/codingstyle.py
```diff
@@ -302,3 +302,60 @@ def feed(self, entry):
...
+        super().__init__(options)
+        self.regexes = []
+        for regexp, repl in self.REGEXPS:
+            self.regexes.append((re.compile(regexp), repl))
+
+    def feed(self, entry):
+        pkg, lines = entry
+        links = defaultdict(list)
+
+        for lineno, line in enumerate(lines):
```
radhermit: You'll probably want enumerate(lines, 1) here so lineno isn't off by one, but I can probably fix that on merge.

mgorny: Done.

mgorny:
Now includes tests as well.

radhermit:
Thanks, merged with some minor changes.

PR #93, opened 2019-07-25
Add a check for packages using .zip snaps instead of .tar.gz

mgorny:


radhermit:
Thanks, merged with minor changes.

PR #94, opened 2019-07-25
ObsoleteUriCheck: fix replacement GitLab URI, improve re readability

mgorny:
Fix GitLab replacement URI to use the full form that enables correct
directory naming.  While GitLab accepts pretty much anything, only
the one true form guarantees reasonable contents.

While at it, use named groups to make the regexps more readable.


radhermit:
Thanks, merged.

PR #96, opened 2019-07-26
Add check for missing dependency on unpackers

mgorny:


src/pkgcheck/checks/metadata_checks.py
```diff
@@ -1044,3 +1044,90 @@ def feed(self, pkg):
...
+    """Check whether package is missing unpacker dependencies."""
+
+    feed_type = base.versioned_feed
+
+    known_results = (MissingUnpackerDep,)
+    required_addons = (addons.UseAddon,)
+
+    unpackers = {
+        '.zip': ('app-arch/unzip',),
+        '.ZIP': ('app-arch/unzip',),
```
Arfrever: `unpack` in EAPI >= 6 matches suffixes case-insensitively.
So these duplicate entries for `.ZIP`, `.7Z`, `.RAR`, `.LHA`, `.LHa` can be dropped, and instead `.lower()` should be used somewhere.

mgorny: Done.

mgorny:
Rebased.

src/pkgcheck/checks/metadata.py
```diff
@@ -1044,3 +1044,85 @@ def feed(self, pkg):
...
+
+class MissingUnpackerDepCheck(base.Template):
+    """Check whether package is missing unpacker dependencies."""
+
+    feed_type = base.versioned_feed
+
+    known_results = (MissingUnpackerDep,)
+    required_addons = (addons.UseAddon,)
+
+    unpackers = {
```
radhermit: I generally try to keep class attrs immutable unless otherwise required, i.e. use an ImmutableDict here.

src/pkgcheck/checks/metadata.py
```diff
@@ -1044,3 +1044,85 @@ def feed(self, pkg):
...
+        fetchables, _ = self.fetch_filter((fetchable,), pkg,
+            pkg._get_attr['fetchables'](
+                pkg, allow_missing_checksums=True,
+                ignore_unknown_mirrors=True, skip_default_mirrors=True))
+
+        # formats is used as definitive list of formats for which deps
+        # are missing (since format may allow multiple unpackers)
+        formats = set()
+        # unpackers are used to optimize dep checks
+        unpackers = set()
```
radhermit: I think changing the name of this or the class attr with the same name would be good.

mgorny:
Updated as requested: used `ImmutableDict`, and added `rem_` prefix to local vars (for 'remaining').

src/pkgcheck/checks/metadata.py
```diff
@@ -1044,3 +1045,85 @@ def feed(self, pkg):
...
+    feed_type = base.versioned_feed
+
+    known_results = (MissingUnpackerDep,)
+    required_addons = (addons.UseAddon,)
+
+    unpackers = ImmutableDict({
+        '.zip': ('app-arch/unzip',),
+        '.jar': ('app-arch/unzip',),
+        '.7z': ('app-arch/p7zip',),
+        '.rar': ('app-arch/rar', 'app-arch/unrar'),
```
radhermit: Do we actually support unpacking via rar? PMS states, "Ebuilds must ensure that RARLAB’s unrar is installed."

src/pkgcheck/checks/metadata.py
```diff
@@ -1044,3 +1045,85 @@ def feed(self, pkg):
...
+        '.lha': ('app-arch/lha'),
```
radhermit: Both the unpacker values for .lha and .lzh aren't tuples but should be to work with the current code.

src/pkgcheck/checks/metadata.py
mgorny: RAR installs `unrar` executable as well, so there's no reason why it wouldn't work. Also, this kind of ambiguity is why I'd rather kill all those extra formats long-term.

src/pkgcheck/checks/metadata.py
mgorny: Good catch. /me mumbles about weird Python syntax.

mgorny:
Fixed.

radhermit:
Thanks, merged with it largely rewritten and simplified along with some additional tests.

PR #97, opened 2019-07-27
BadRestrictsCheck: permit preserve-libs

mgorny:
RESTRICT=preserve-libs has been added to Portage recently.  Since
the use case seems valid, and PMS permits additional values not listed
in the spec, stop complaining about it.

// NB: I'm not including network-sandbox since it's not allowed in ::gentoo.


radhermit:
I'd rather drop the check then add restricts that portage implements, otherwise where does it stop? Since the spec doesn't appear to have any limitations on the format of the tokens in RESTRICT it's probably not worth checking at all.

mgorny:
The alternative is to explicitly complain about restricts that are forbidden in gentoo.

radhermit:
Document the official list somewhere like the devmanual/wiki or dump it in the tree and then I'll be fine with supporting this check with tree-approved updates. Otherwise it seems too random, imo.

PR #98, opened 2019-07-29
Add a check for combining two paths with embedded EPREFIX

mgorny:
// NB: there's also a need for similar check for using `doins $(python_get_sitedir)` etc.

### DIFF
--- a/src/pkgcheck/checks/codingstyle.py
+++ b/src/pkgcheck/checks/codingstyle.py
@@ -167,30 +167,94 @@ class UnnecessarySlashStrip(base.Warning):
         return f"{self.variable} unnecessary slash strip on line{_pl(self.lines)}: {lines}"
 
 
+class DoublePrefixInPath(base.Error):
+    """Ebuild uses two consecutive paths including EPREFIX.
+
+    Ebuild combines two path variables (or a variable and a getter), both
+    of which include EPREFIX, resulting in double prefixing.  This is the case
+    when combining many pkg-config-based or alike getters with ED or EROOT.
+
+    For example, '${ED}$(python_get_sitedir)' should be replaced
+    with '${D}$(python_get_sitedir)'.
+    """
+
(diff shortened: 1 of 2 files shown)


src/pkgcheck/checks/codingstyle.py
```diff
@@ -203,6 +267,9 @@ def feed(self, entry):
             yield MissingSlash(pkg, var, lines)
         for var, lines in unnecessary.items():
             yield UnnecessarySlashStrip(pkg, var, lines)
+        print(double_prefix)
```
radhermit: I imagine this is a errant debugging line.

mgorny: Sorry, removed now.

src/pkgcheck/checks/codingstyle.py
```diff
@@ -167,31 +167,95 @@ def short_desc(self):
...
 class PathVariablesCheck(base.Template):
     """Scan ebuild for path variables with various issues."""
 
     feed_type = base.ebuild_feed
-    known_results = (MissingSlash, UnnecessarySlashStrip)
-    variables = ('ROOT', 'EROOT', 'D', 'ED')
+    known_results = (MissingSlash, UnnecessarySlashStrip, DoublePrefixInPath)
+    prefixed_variables = ('EROOT', 'ED')
+    variables = ('ROOT', 'D') + prefixed_variables
+    prefixed_getters = (
```
radhermit: Would it be possible to flag these in a formatted way somehow using the eclass doc markup so we don't have to maintain the list manually in pkgcheck once eclass doc parsing support is written?

mgorny: Hmm, I don't see a problem with that. I suppose an extra tag for VARIABLE and FUNCTION thingies.

radhermit:
Merged with a TODO note mentioning the above. Once eclass parsing is done, I'll return to poke you/whoever to fix that if it's not done by then. :smile:

PR #100, opened 2019-07-29
metadata: Improve documentation for common reports

mgorny:
Improve the documentation to explicitly state what is wrong, and how it
can be fixed.

### DIFF
--- a/src/pkgcheck/checks/metadata.py
+++ b/src/pkgcheck/checks/metadata.py
@@ -140,7 +140,17 @@ class MetadataCheck(base.Template):
 
 
 class RequiredUseDefaults(base.Warning):
-    """Default USE flag settings don't satisfy REQUIRED_USE."""
+    """Default USE flag settings don't satisfy REQUIRED_USE.
+
+    The REQUIRED_USE constraints specified in the ebuild are not satisfied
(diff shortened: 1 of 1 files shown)


src/pkgcheck/checks/metadata.py
```diff
@@ -347,7 +368,20 @@ def feed(self, pkgs):
...
+    """Missing slot value in dependencies.
+
+    The package dependency does not specify a slot but the target package
+    has multiple slots. The behavior for satisfying this kind of dependency
+    is not strictly defined, and may result in either any or the newest package
+    slot being accepted.
+
+    Please verify whether the package works with all the dependency slots.
+    If only one slot is actually acceptable, specify it explicitly. If multiple
+    slots are acceptable, please use either ':=' or explicit ':*' slot operator.
```
radhermit: One thing to note that I'll fix locally while merging, be careful using text that will be interpreted as reStructuredText markup, in this case the '*' char. I generally make them string literals where it makes sense to (so '``:*``' in this case) or otherwise they need to be escaped so the lineflow isn't broken for the generated docs.

radhermit:
Thanks, merged with minor rst formatting changes.

PR #102, opened 2019-07-30
python: Fix false positive reports due to blocker atoms

mgorny:


mgorny:
Fixed the test case that didn't trigger properly before.

radhermit:
Thanks, merged.

PR #108, opened 2019-08-03
 Add checks for acct-* package identifiers

mgorny:


src/pkgcheck/checks/acct.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def __init__(self, options):
+        super().__init__(options)
+        self.id_re = re.compile(
+            r'ACCT_(?P<var>USER|GROUP)_ID=(?P<quot>[\'"]?)(?P<id>[0-9]+)(?P=quot)')
+        self.seen_uids = defaultdict(lambda: defaultdict(list))
+        self.seen_gids = defaultdict(lambda: defaultdict(list))
+
+    def feed(self, full_pkgset):
+        # TODO: can we filter this earlier?
```
mgorny: Note this TODO. The check would be much faster if it didn't have to skip all the other packages explicitly.

radhermit: At present, it would require writing a custom source class that would feed only what the check requested. That shouldn't be too hard since I recently extended pkgcheck to add a filtered source (respects package.mask) and a git commits source (only yields pkgs that have been changed locally in git). It would also make me have to figure out how to merge source results back into a sane order (which I was saving for parallelization work when the issue is much more prevalent.)

mgorny:
Updated UID/GID ranges.

mgorny:
Rebased.

radhermit:
Thanks, merged and then reworked in f1b37e5ddf4dafac64b41b77e1b4a357712ecd98.

PR #110, opened 2019-08-03
Add a check for correct maintainer types

mgorny:
Add a check that reports project-type maintainers that are not present
in projects.xml, and person-type maintainers that match projects.

// Requires pkgcore/pkgcore#265 and pkgcore/pkgcore#267


radhermit:
Thanks, merged with some changes in the commit after it.

PR #113, opened 2019-08-04
python: Exempt virtual/pypy* from eclass check

mgorny:
The virtual/pypy* is used as an implementation package, and its special
contents trigger false positive.  While I could technically try to hack
it over, it's probably cleaner to exempt it straight in the check.

### DIFF
--- a/src/pkgcheck/checks/python.py
+++ b/src/pkgcheck/checks/python.py
@@ -21,6 +21,8 @@ INTERPRETERS = frozenset([
     'virtual/pypy3',
 ])
 
+CHECK_EXCLUDE = frozenset(['virtual/pypy', 'virtual/pypy3'])
+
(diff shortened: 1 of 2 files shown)


src/pkgcheck/checks/python.py
```diff
@@ -237,6 +237,11 @@ def feed(self, pkg):
             return
 
         if eclass is None:
+            # virtual/pypy* need to be exempted as they serve as slot-matchers
+            # for other packages
+            if pkg.key in ('virtual/pypy', 'virtual/pypy3'):
```
radhermit: This tuple seems like it should be a class attr frozenset named something apt since more values could be added in the future.

mgorny: Added it to the global vars on top.

radhermit:
Thanks, merged.

PR #115, opened 2019-08-06
Improve message & doc for ObsoleteUri

### LINKED ISSUE #114: ObsoleteURI check needs to provide more help (or documentation)
Especially with the github URI's, moving from one to the other causes a checksum to change, which introduces fun problems if the old version is cached on distfiles.gentoo.org:

- Maintainer-nuking `Manifest` and `/usr/portage/distfiles` and running `repoman manifest` gives you a wrong `Manifest` entry if the user fetches directly from the URL mentioned in `SRC_URI` (due to default mechanics being to simply refetech from `distfiles.gentoo.org`)
- running `repoman manifest` with `GENTOO_MIRRORS="" repoman manifest` gets you the right checksums, but at the expense that users will still fetch from the `distfiles.gentoo.org` mirror, and get wrong checksummed files
- In some cases, this specific URI might suggest maintainers look and see if upstream have a `/releases/` page, and consider using those results instead, however, doing so is still subject to the original two problems above.
- Adding a suffix in `SRC_URI` like ` -> ${P}.gh2.tar.gz` will satisfy the aforementioned checksum issues, but will now create a problem in `src_prepare` due to the tarball being extracted as `${WORKDIR}/${P}.gh2` and `src_prepare` will try to  access `${WORKDIR}/${P}` which doesn't exist, so you *also* need to make sure you do `S="${WORKDIR}/${P}.gh2"`

In the case above of "migrate from `/tarball/` to `/releases/` instead of `/archive/`, I humbly  suggest using the suffix `.original.tar.gz` ( as its the original sources provided by upstream, not some mess of github rolling a tar from a SHA1 with risks of random checksum changes )

I suggest some documentation be at least, placed here: https://pkgcheck.readthedocs.io/en/latest/man/pkgcheck.html#obsoleteuri , but some mention of some kind of pitfall aught to be mentioned in the warning itself.

Enhancing the error in any way that leads developers to the best conclusion is greatly desired, especially as this is the sort of situation where somebody could think "Oh, easy fix", test it, find no bugs, and only later work it out after much confusion from a cryptic user bug report.

( I would hope, for a minimum effort approach, maintainers might see this bug before trusting `repoman commit` to make sure this mess doesn't happen... )

mgorny:
Fixes #114

### DIFF
--- a/src/pkgcheck/checks/codingstyle.py
+++ b/src/pkgcheck/checks/codingstyle.py
@@ -377,7 +377,10 @@ class ObsoleteUri(base.Warning):
     """URI used is obsolete.
 
     The URI used to fetch distfile is obsolete and can be replaced
-    by something more modern.
+    by something more modern. Note that the modern replacement usually
+    results in different file contents, so you need to rename it (to
+    avoid mirror collisions with the old file) and update the ebuild
+    (for example, by removing no longer necessary vcs-snapshot.eclass).
     """
 
     __slots__ = ("category", "package", "version", "line", "uri", "replacement")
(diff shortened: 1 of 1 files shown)


radhermit:
Thanks, merged the docstring update and dropped the additional message component.

PR #117, opened 2019-08-07
Split invalid/old copyright checks outta git checks, and improve

### LINKED ISSUE #116: Split copyright checks into git and non-git part
While git part is understandably beneficial to checking dates and so on, there are checks that could be performed without having git checkout. In particular, `InvalidCopyright` and `OldGentooCopyright` (based on switch date).

mgorny:
Split checks for invalid copyright lines and 'Gentoo Foundation'
copyright outside the git checks since they can really work without git
and are quite useful.

Improve the regex used to catch copyrights to grab dates and copyright
holder.  Use the latter to detect 'Gentoo Foundation' rather than
a second regex match.  Furthermore, limit catching GF to ebuilds
committed in 2019 and newer, according to the copyright date.
If ebuilds have wrong copyright date, OutdatedCopyright should catch
that.

Also improve documentation to clearly indicate how copyright lines
should look like, and what holder to use for various repos.

Fixes #116

### DIFF
--- a/src/pkgcheck/checks/codingstyle.py
+++ b/src/pkgcheck/checks/codingstyle.py
@@ -487,11 +487,34 @@ class OldGentooCopyright(base.Warning):
         return f'old copyright, update to "Gentoo Authors": {self.line!r}'
 
 
+class NonGentooAuthorsCopyright(base.Warning):
+    """Ebuild with copyright stating owner other than "Gentoo Authors".
+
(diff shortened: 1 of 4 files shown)


mgorny:
Also added a check for not using 'Gentoo Authors' in ::gentoo.

radhermit:
Thanks, merged with some basic changes and made the check exclusive to the gentoo repo. External repos/overlays don't even need to use copyrights if they don't care to.

PR #119, opened 2019-08-11
Check for prefixed paths combined with *into, dodir, fperms, fowners

mgorny:
This is quite fragile but it catches a lot of mistakes anyway.  I've
been able to kill all but one false positives from ::gentoo, and the one
left most likely suffers from indirect EPREFIX output anyway.


radhermit:
Thanks, merged.

PR #121, opened 2019-08-13
Add check for using ${HOMEPAGE} in SRC_URI

mgorny:


radhermit:
Thanks, merged with minor changes.

orlitzky:
Using git/pkgcheck to replace repoman is a much better workflow, but I'm never going to use it if it warns me about things that aren't wrong and that I'm not going to fix (like this).

radhermit:
> Using git/pkgcheck to replace repoman is a much better workflow, but I'm never going to use it if it warns me about things that aren't wrong and that I'm not going to fix (like this).

You can easily disable specific checks and keywords from being scanned for, but really I don't care what you use and this really wasn't a great way to complain about things you don't like.

ferringb:
@orlitzky a suggestion- you're assuming everyone knows what you mean by "things that aren't wrong".  That's a wrong assumption, and that's assuming you're right.

Clarify please- there are ways to build configurables here to vary behaviour, but you're not providing any form of constructive (or actionable) criticism,

orlitzky:
When this check was introduced, the CI tool that runs on ::gentoo started sending out emails stating that the repository is broken (**WARNING: repository still has warnings!**). It's not. The end result is that I now ignore any emails I get from the CI tool, because I'm not going to read through a mile-long report every time I make a commit just to find out that my "mistake" is putting a variable in `SRC_URI`. This obviously isn't great, and it's why projects like GCC and clang are careful about which warnings they enable by default. But in any case, the CI tool isn't pkgcheck's fault.

Inspired by a mailing list thread, mgorny made a [blog post describing an improved workflow using pkgcheck/git](https://blogs.gentoo.org/mgorny/2019/12/12/a-better-ebuild-workflow-with-pure-git-and-pkgcheck/). The workflow is better, but to automate it, I have to create a custom configuration that hides warnings from pkgcheck. And that's no good, because we don't want developers ignoring warnings before committing.

So why should I be trying to ignore *this* warning? Because it doesn't indicate that anything is wrong, and doesn't need to be changed. No one is hurt by it, and nothing is improved by copy/pasting the value of `${HOMEPAGE}` into the spot in `SRC_URI` where it says `${HOMEPAGE}`. It never should have been added as a warning in the first place.

For actionable criticism: this should be an informational message if anything, and not a warning. Even then, it's less useful than most informational messages. If there was a class of messages that were enabled by whitelisting instead of by blacklisting, then that would be a better place for it instead.

radhermit:
Responding to a few of the pkgcheck related things: currently pkgcheck operates as many linting tools seem to, e.g. pylint, in that almost all results are enabled and you can disable them per target or scan for exactly selected checks/results. I generally prefer that route while it can be verbose since it forces visibility for most new results instead of hiding them, forcing users to actually read the docs/changes to enable them (which most won't do).

In regards to "creating a custom configuration", that currently requires a one line shell alias, so I'm not sure how hard that is to set up properly. However, I'd like to support a custom configuration for pkgcheck that allows per repo settings more easily and there should already be an issue open for that.

Finally, with respect to this check in particular the main reason is described in the [docstring for the result](https://pkgcheck.readthedocs.io/en/latest/man/pkgcheck.html#homepageinsrcuri), specifically that HOMEPAGE is designed to have multiple values and changing it shouldn't always affect SRC_URI. Whether you think that's important enough for flagging at a specific level is something to question QA or related people about.

orlitzky:
> Whether you think that's important enough for flagging at a specific level is something to question QA or related people about.

That was my aim here, but I seem to have missed up and to the left a bit and hit you and ferringb instead :santa:

PR #122, opened 2019-08-14
MissingUseDepDefault: fix var name in constructor

mgorny:


radhermit:
Thanks, merged with initial tests for MissingUseDepDefault.

PR #123, opened 2019-08-15
DoublePrefixInPath: improve message, variables → paths

mgorny:
Change the DoublePrefixInPath message to explicitly say we're
concatenating paths, not necessarily variables.  This should avoid
confusion now that it detect *into calls.


radhermit:
Thanks, merged.

PR #138, opened 2019-09-04
Add a test for USE flags with reserved underscore character

mgorny:
@radhermit, used yer fancy new test framework.

CC @ulm.


radhermit:
Thanks, merged.

PR #142, opened 2019-09-11
reporters: Add CsvReporter, convenient for shell pipelines

mgorny:


radhermit:
Tests are missing and should be easy to add.

mgorny:
Sorry, didn't occur to me you have tests for that ;-). Added them, though I suppose it'd be nice to use some result with `,` to test quoting.

radhermit:
Thanks, merged with altered results to include commas for quote testing.

PR #181, opened 2019-12-01
git: Read .git/info/exclude as well as .gitignore

chutz:
Git will read "ignore" entries from both `.gitignore` and `.git/info/exclude`.
We should do the same thing so we have the same behaviour.


radhermit:
Merged with minor changes.

PR #192, opened 2019-12-25
DependencyCheck: do not report DeprecatedPkg for blockers

mgorny:
Reported by gyakovlev.

### DIFF
--- a/src/pkgcheck/checks/metadata.py
+++ b/src/pkgcheck/checks/metadata.py
@@ -747,7 +747,10 @@ class DependencyCheck(Check):
                     in_or_restriction = False
 
                 for atom in iflatten_instance(node, (atom_cls,)):
-                    if self.deprecated(atom):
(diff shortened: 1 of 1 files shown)


src/pkgcheck/checks/metadata.py
```diff
@@ -747,7 +747,7 @@ def feed(self, pkg):
                     in_or_restriction = False
 
                 for atom in iflatten_instance(node, (atom_cls,)):
-                    if self.deprecated(atom):
+                    if not atom.blocks and self.deprecated(atom):
```
ferringb: You might want to add a comment here explaining 'why'- it took me a bit to think through why you were adding this.  I agree with it- at first glance- but a comment would've made the understanding quicker, including for folks who look at this down the line.

mgorny: Done.

Issue #186, opened 2019-12-12
git: implement more GLEP 66 checks


mgorny:
Since pkgcheck is now checking commit message summary prefix, it'd be nice to integrate some more GLEP 66 checks that are present in repoman:

- making sure that summary is a single line,
- checking that it fits <= 69 chars (repoman also accepts pkg: + 50 chars in case pkg was very long),
- checking that body is wrapped at 80 chars (with exception if body has only one word -- then it's probably URI or some identifier we can't wrap),
- making sure there's only one footer (i.e. that all footer tags occur in one block, without empty lines in between),
- forbidding `Gentoo-Bug`,
- making sure `Bug` and `Closes` specify URL, not number.

One thing missing in repoman but still would be useful:
- making sure `Fixes` is used to reference commit id, not bug URL or number (in infra hooks we check whether it starts with at least 8 hex digits ­— but I suppose you could even verify whether it's in git history).

radhermit:
I'd like to figure out a sane way to add proper git check and result testing before adding too much additional git-related support so I'll probably try to do that work along with this type of request.

juippis:
- Check for **proper** sign-off syntax. There are currently some PRs with sign-offs missing `< >` around e-mail addresses.

radhermit:
> * Check for **proper** sign-off syntax. There are currently some PRs with sign-offs missing `< >` around e-mail addresses.

Just to note, none of these checks really matter atm in relation to github PRs since afaik contributors are still pointed to use repoman by docs and these git-related checks are only run on local systems when --commits is used, i.e. not by CI.

mgorny:
But we do use it to verify their commits ;-).

radhermit:
> But we do use it to verify their commits ;-).

Fair enough, in that case I'd also point out that these checks should be some of the easiest to submit as a PR since all that's mostly done is various mangling/regexing of the commit.message attribute.

radhermit:
I think most of these are done now except for some of the tag specific stuff.

radhermit:
> * Check for **proper** sign-off syntax. There are currently some PRs with sign-offs missing `< >` around e-mail addresses.

This should already be working since git enforces 'name <email>' format for author/committer fields and pkgcheck uses those to match against. I don't think it's worth reverifying that format if git is already handling it.

radhermit:
[...]

PR #193, opened 2019-12-26
Bug/Closes commit validation and initial tests

ferringb:
Commit messages detail things more fully; this is an initial stab at #186 , and adding tests for `MissingSignOff` while I'm at it.

This is a fairly simple bit of logic- I'm using this as a way to edge back into the code including a forum for any comments/suggestions for aligning my coding style w/ what the modern pkgcore version is.  I'm assuming I'm in the range of it, but if not- please nitpick the hell out of it.

### DIFF
--- a/src/pkgcheck/checks/git.py
+++ b/src/pkgcheck/checks/git.py
@@ -5,6 +5,7 @@ import tarfile
 from datetime import datetime
 from itertools import chain
 from tempfile import TemporaryDirectory
+from urllib.parse import urlparse
 
 from pkgcore.ebuild.misc import sort_keywords
 from pkgcore.ebuild.repository import UnconfiguredTree
@@ -318,19 +319,54 @@ class MissingSignOff(results.CommitResult, results.Error):
         return f'commit {self.commit}, missing sign-off{s}: {sign_offs}'
 
 
+class InvalidTagFormat(results.CommitResult, results.Error):
+    """Local commit has a tag that is incompliant.
+
+    Commit tags have restrictions as to the allowed format and data
+    used per GLEP 66 [#]_.
+
+    .. [#] https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin
+    """
+
+    def __init__(self, tag, value, error, **kwargs):
+        super().__init__(**kwargs)
+        self.tag, self.value, self.error = tag, value, error
+
+    @property
+    def desc(self):
+        return f'commit {self.commit}, tag {self.tag} value {self.value}: {self.error}'
+
+
 class GitCommitsCheck(GentooRepoCheck, ExplicitlyEnabledCheck):
     """Check unpushed git commits for various issues."""
 
     scope = base.commit_scope
     _source = GitCommitsSource
-    known_results = frozenset([MissingSignOff])
+    known_results = frozenset([MissingSignOff, InvalidTagFormat])
 
     def feed(self, commit):
-        # check for missing git sign offs
+        # check for missing git sign offs-
         sign_offs = {
-            line[15:].strip() for line in commit.message
-            if line.startswith('Signed-off-by: ')}
+           line[15:].strip() for line in commit.message
+           if line.startswith('Signed-off-by: ')}
         required_sign_offs = {commit.author, commit.committer}
         missing_sign_offs = required_sign_offs - sign_offs
         if missing_sign_offs:
             yield MissingSignOff(tuple(sorted(missing_sign_offs)), commit=commit)
+
+        url_tags = (x.split(": ", 1) for x in commit.message)
+        url_tags = [
+            (x[0], x[1])
+            for x in url_tags if len(x) == 2 and x[0] in ('Bug', 'Closes')
+        ]
+        # check tag formats.
+        for tag, value in url_tags:
+            parsed = urlparse(value)
+            if not parsed.scheme:
+                yield InvalidTagFormat(tag, value, "value isn't a URL", commit=commit)
+                continue
+            if parsed.scheme.lower() not in ("http", "https"):
+                yield InvalidTagFormat(
+                    tag, value,
+                    "Invalid protocol; should be http or https",
+                    commit=commit)
--- a/src/pkgcheck/checks/git.py
+++ b/src/pkgcheck/checks/git.py
@@ -23,6 +23,10 @@ demand_compile_regexp(
     'ebuild_copyright_regex',
     r'^# Copyright (\d\d\d\d(-\d\d\d\d)?) .+')
 
+demand_compile_regexp(
+    'commit_footer',
+    r'^(?P<tag>[a-zA-Z0-9_-]+): (?P<value>.*)$')
+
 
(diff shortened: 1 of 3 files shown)


tests/module/checks/test_git.py
```diff
@@ -0,0 +1,75 @@
+from pkgcheck.checks import git as git_mod
+from .. import misc
+from pkgcheck.git import _GitCommit
+
+def mk_commit(
```
ferringb: @radhermit if you've got a preferred style for this sort of indentation, let me know.  I'm following a rough version of google norms- frankly more out of habit then preference.

tests/module/checks/test_git.py
```diff
@@ -0,0 +1,75 @@
...
+            self.check,
+            (self.commit("blah", author="user1", committer="user2"),)
+        )
+        assert isinstance(r, git_mod.MissingSignOff)
+        assert r.missing_sign_offs == ('user1', 'user2')
+
+        # assert that it handles author/committer being the same
+        self.assertNoReport(
+            self.check,
+            (self.commit("Signed-off-by: user@user.com",
```
ferringb: Remember that commit objects are just NamedTuples- this interacts poorly with the machinery in `misc.ReportTestCase`, hence having to do the manual tuple wrapping here.

One way to eliminate that- create actual classes for `_GitCommit`.  Doing so would also allow caching parsed tags- doing the parse once.  I'd probably implement it (since I'm usually API orientated), but it's your call if this is actually useful or not- I've been out of this codebase for many, many years.

Let me know your preference-

ferringb:
Regarding test failures- local tox run (same for pytest) pass.  The complaint centers on the importation of pkgcheck.git._GitCommit- which does exist.  Any pointers for why tox/travis is complaining here?

radhermit:
> Regarding test failures- local tox run (same for pytest) pass. The complaint centers on the importation of pkgcheck.git._GitCommit- which does exist. Any pointers for why tox/travis is complaining here?

It looks like you're not working against git head. pkgcheck.git._GitCommit got replaced in 7b4389de567397569acfd247f89e792039a4c85b.

tests/module/checks/test_git.py
radhermit: Latest pkgcheck does use NamedTuple classes from the typing module for GitCommit. Does that help out in this case?

tests/module/checks/test_git.py
radhermit: It mostly looks fine, although I usually try not to put the closing parenthesis on the same indent level as the function-scoped code.

tests/module/checks/test_git.py
```diff
@@ -0,0 +1,50 @@
...
+    check_kls = git_mod.GitCommitsCheck
+    check = git_mod.GitCommitsCheck(None)
+
+    @staticmethod
+    def commit(message, **kwargs):
+        """Helper function to create commits and cleanup message
+
+        Due to developer convenience, errant leading whitespace is
+        in the messages- this will strip that out so tags are properly
+        parsed."""
```
radhermit: Generally prefer docstring end marks to be on their own line.

radhermit: For multiline docstrings, that is.

ferringb: Ack- works for me.

tests/module/checks/test_git.py
```diff
@@ -0,0 +1,50 @@
...
+            self.check,
+            (self.commit("Signed-off-by: user@user.com",
+                author="user@user.com", committer="user@user.com"),
+            )
+        )
+
+        # assert it can handle multiple sign offs.
+        self.assertNoReport(
+            self.check,
+            (self.commit("Signed-off-by: user2\nSigned-off-by: user1",
```
radhermit: Continued indent level usage with parenthesis is usually handled closer to what pylint suggests, for example:

```
(self.commit(
    "Signed-off-by: user2\nSigned-off-by: user1",
     author="user1", committer="user2"),
)
```

tests/module/checks/test_git.py
ferringb: Nope- it's still a tuple behind the scenes, the collections.NamedTuple (and types) is just wrappers of the class injecting some extra things.  types.NamedTuple is a nice change however.

src/pkgcheck/checks/git.py
```diff
@@ -318,19 +319,54 @@ def desc(self):
...
+    .. [#] https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin
+    """
+
+    def __init__(self, tag, value, error, **kwargs):
+        super().__init__(**kwargs)
+        self.tag, self.value, self.error = tag, value, error
+
+    @property
+    def desc(self):
+        return f'commit {self.commit}, tag {self.tag} value {self.value}: {self.complaint}'
```
radhermit: Don't you mean to use self.error here instead of self.complaint?

tests/module/checks/test_git.py
```diff
@@ -0,0 +1,75 @@
...
+        r = self.assertReport(self.check, f("Bug: 123455"))
+        assert isinstance(r, git_mod.InvalidTagFormat)
+        assert (r.tag, r.value, r.error) == ('Bug', '123455', "value isn't a URL")
+
+        # do a protocol check; this is more of an assertion against the parsing model
+        # used in the implementation.
+        r = self.assertReport(self.check, f("Closes: ftp://blah.com/asdf"))
+        assert isinstance(r, git_mod.InvalidTagFormat)
+        assert r.tag == 'Closes'
+        assert "protocol" in r.error
```
radhermit: I'd assert directly against the str(r) instead of r.error so you can catch attribute/result encoding issues like the complaint/error one I mentioned above.

tests/module/checks/test_git.py
ferringb: ack, will make that change shortly.

src/pkgcheck/checks/git.py
ferringb: That I do.  I'm going to modify the test misc.assertReportSanity so it forces a str() of the report to catch things like this in addition to fixing what I screwed up here ;)

tests/module/checks/test_git.py
radhermit: Actual class objs would be fine, I mostly used named tuples since I thought they'd be simpler and possibly help full git trees with grafted history get parsed faster on initial runs without resorting to dropping into CPython/cython.

However, I also haven't done any benchmarking to know how much the difference actually is, but I'd guess probably not that much in this case.

tests/module/checks/test_git.py
ferringb: I've added an assertion that rendering succeeds- in this case, I'd prefer to decouple the exact error message from the text.  In this case, we really only care that someone didn't slip in some non http protocol.

radhermit:
Thanks, merged with a bunch of fixes/changes on top.

ferringb:
@radhermit appreciated, and pardon the lag- holidays kind of nuked my keyboard time.

PR #210, opened 2020-02-07
Support for updated python-single-r1.eclass

mgorny:
Note: I expect tests to fail, I'm abusing your CI ;-).

### DIFF
--- a/src/pkgcheck/checks/python.py
+++ b/src/pkgcheck/checks/python.py
@@ -48,31 +48,6 @@ class MissingPythonEclass(results.VersionResult, results.Warning):
         return f'missing {self.eclass} eclass usage for {self.dep_type}="{self.dep}"'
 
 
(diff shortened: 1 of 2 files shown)


radhermit:
> Note: I expect tests to fail, I'm abusing your CI ;-).

That's fine, but I don't think I've enabled CI for PRs yet. I'll try to do that today.

mgorny:
It's kinda curious because I got some CI mail about failure. But maybe it's getting enabled automatically on my fork or w/e.

radhermit:
Now it should if you re-push.

mgorny:
Ok, it seems to be good now. If you could release it ASAP, that would be awesome. This is needed for gentoo/gentoo#14589.

radhermit:
Thanks, merged.

Also, at some point it would be nice if you could add ebuilds that trigger the related results to a repo in tests/repos along with the expected output and patches to fix them.

Issue #52, opened 2017-04-11
add API support for running checks


radhermit:
It could be nice to split out certain functionality from the main script and make the test process easier to kick off via API calls. This would allow external python scripts/tools to more easily use pkgcheck without having to run it via a script and then parse the output from whatever format is selected.

PR #211, opened 2020-02-15
FormatReporter: add {class} and print empty string for missing attributes

lucianposton:
Two commits that:
- Add {class} to FormatReporter attributes
- Print the empty string for missing --format attributes

### DIFF
--- a/src/pkgcheck/reporters.py
+++ b/src/pkgcheck/reporters.py
@@ -364,6 +364,7 @@ class FormatReporter(Reporter):
             result = (yield)
             attrs = vars(result)
             attrs.update((k, getattr(result, k)) for k in self._properties)
+            attrs['class'] = result.__class__.__name__
             self.out.write(self.formatter.format(self.format_str, **attrs))
             self.out.stream.flush()
 
--- a/src/pkgcheck/reporters.py
+++ b/src/pkgcheck/reporters.py
@@ -8,6 +8,7 @@ import signal
 from collections import defaultdict
 from itertools import chain
 from multiprocessing import Process, SimpleQueue
+from string import Formatter
 from xml.sax.saxutils import escape as xml_escape
 
 from snakeoil import pickling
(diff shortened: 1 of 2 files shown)


radhermit:
Output name instead of class has been added to available formatter output properties in https://github.com/pkgcore/pkgcheck/commit/e1e4de36ab0683e344f87f993eda2bf30d87a5c3.

In terms of using the empty string for missing --format attributes you'll have to make your case and explain why it's needed.

lucianposton:
> In terms of using the empty string for missing --format attributes you'll have to make your case and explain why it's needed.

@radhermit  It's been awhile since I looked into this, but if I recall correctly, if the --format specifies an attribute and a result is missing a value for that attribute, nothing at all is output for that result. The expected behavior is that the non-empty attributes for that result are output since the user specified them via --format, rather than the entire result disappearing as if there were no issue detected.

The use case was writing a tool to automatically run pkgcheck that parsed every attribute, but when one attribute was missing, the entire result was absent from the pkgcheck output.

radhermit:
> @radhermit It's been awhile since I looked into this, but if I recall correctly, if the --format specifies an attribute and a result is missing a value for that attribute, nothing at all is output for that result. The expected behavior is that the non-empty attributes for that result are output since the user specified them via --format, rather than the entire result disappearing as if there were no issue detected.

I partially agree with this, but would squelch output for any formatted string consisting of the empty string or all whitespace since those cases seem nearly worthless to output for.

> The use case was writing a tool to automatically run pkgcheck that parsed every attribute, but when one attribute was missing, the entire result was absent from the pkgcheck output.

If this tool was written in python I imagine this use case would be better covered using the API directly once work for https://github.com/pkgcore/pkgcheck/issues/52 was finished. Otherwise for a different language, I'd think an entirely different reporter would work better if you want a mapping between existing attribute names and their values for each result.

Issue #212, opened 2020-02-15
Crash when maintainer email is empty


lucianposton:
pkgcheck is crashing from an unhandled exception from pkgcore due to an empty maintainer email/name in metadata.xml, such as the default one created by app-vim/gentoo-syntax.

```
Traceback (most recent call last):
  File "/usr/lib64/python3.6/site-packages/pkgcheck/pipeline.py", line 71, in _run_checks
    results.extend(pipe.run(restrict))
  File "/usr/lib64/python3.6/site-packages/pkgcheck/pipeline.py", line 189, in run
    yield from check.feed(item)
  File "/usr/lib64/python3.6/site-packages/pkgcheck/checks/metadata.py", line 472, in feed
    local_use = pkg.local_use
  File "/usr/lib64/python3.6/site-packages/snakeoil/klass.py", line 373, in __call__
    return self.getter(obj)
  File "/usr/lib64/python3.6/site-packages/snakeoil/currying.py", line 71, in callit
    return func(*(moreargs + args), **morekwargs)
  File "/usr/lib64/python3.6/site-packages/pkgcore/ebuild/repo_objs.py", line 95, in _generic_attr
    self._parse_xml()
  File "/usr/lib64/python3.6/site-packages/pkgcore/ebuild/repo_objs.py", line 163, in _parse_xml
    MetadataXml._parse_xml(self, open(self._source, "rb", 32768))
  File "/usr/lib64/python3.6/site-packages/pkgcore/ebuild/repo_objs.py", line 128, in _parse_xml
    maint_type=x.get('type')))
  File "/usr/lib64/python3.6/site-packages/pkgcore/ebuild/repo_objs.py", line 59, in __init__
    raise ValueError('need at least one of name and email')
ValueError: need at least one of name and email
```

radhermit:
Should have been fixed in https://github.com/pkgcore/pkgcore/commit/e1a67fb68018dda6dd0db0ae808d697a760eb38c as mentioned in the related pr.

PR #213, opened 2020-02-15
test case and fix for empty maintainer email

lucianposton:
Two commits:
- test case to repr #212 
- fix for above, empty maintainer email

### DIFF
--- a/src/pkgcheck/pipeline.py
+++ b/src/pkgcheck/pipeline.py
@@ -189,6 +189,9 @@ class CheckRunner:
                     yield from check.feed(item)
                 except MetadataException as e:
                     self._metadata_error_cb(e)
+                except ValueError:
+                    # These pkgcore errors are reported as PkgInvalidXml issues
+                    pass
             self._running_check = None
 
         while self._metadata_errors:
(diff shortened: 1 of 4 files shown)


lucianposton:
```
>                           assert self._render_results(
                                results, verbosity=verbosity) == expected
E                           AssertionError: assert 'PackageMetad...t expected.\n' == 'PackageMetad...t expected.\n'
E                               PackageMetadataXmlCheck/PkgInvalidXml
E                                 PkgInvalidXml: package metadata.xml violates metadata.xsd:
E                               line 5, col 0: (SCHEMAV_CVC_PATTERN_VALID) Element 'email': [facet 'pattern'] The value '' is not accepted by the pattern '[^@]+@[^.]+\..+'.
E                             + line 5, col 0: (SCHEMAV_CVC_DATATYPE_VALID_1_2_1) Element 'email': '' is not a valid value of the atomic type 'emailType'.
E                             - line 5, col 0: (SCHEMAV_CVC_IDC) Element 'email': Warning: No precomputed value available, the value was either invalid or something strange happened.
E                             ?                                                        ...
E                             
E                             ...Full output truncated (4 lines hidden), use '-vv' to show

../../tests/module/test_pkgcheck.py:423: AssertionError
```

fyi, the test runner uses a different dtd than what's used when the tests are run locally via `python setup.py test`, it would seem.

radhermit:
It's better to handle this in pkgcore itself since we generally want lower-level exceptions such as ValueError to propagate and cause obvious crashes/backtraces since they're much too generic to know where they specifically are arising from when running the check pipeline.

Anyway, this should have been fixed via https://github.com/pkgcore/pkgcore/commit/e1a67fb68018dda6dd0db0ae808d697a760eb38c a while ago.

PR #216, opened 2020-02-25
Include PGIDs in reports

mgorny:
Include Policy Guide identifiers in reports. This is going to make it easier to refer to documentation. Long term I also plan to introduce `policy.conf` file to control checks from the repo.


radhermit:
If something like this were to happen, I'd rather move to using some similar form of IDs everywhere similar to how most linters operate instead of using class names as is currently done.

Issue #74, opened 2019-03-13
Check for eclassdoc syntax


mgorny:
Could you include eclassdoc syntax checks, please? I'm particularly thinking of issues that cause eclass-manpages build to fail.

radhermit:
Link to eclass doc formatting reference in the devmanual: https://devmanual.gentoo.org/eclass-writing

Support for parsing this should probably be added to the pkgcore side so it possible could be used for other scripts more easily.

radhermit:
Initial implementation for eclass doc parsing, checks, and cache in the eclass branch. After playing around with all types of python parsing libs and solutions, I just found it easier to roll a manual, regex-based parser to make error reporting easier and more relevant in order to throw proper results for the related check.

radhermit:
The parsing support might be moved to the pkgcore side later as mentioned earlier once support matures in order to support doc generation, scripting, and related tasks targeting eclasses.

PR #238, opened 2020-08-15
DeprecatedEclassCheck: mark cmake-utils as deprecated

sbraz:
Hi, it looks like that message is missing despite being present in repoman, see https://github.com/gentoo/portage/pull/554.

### DIFF
--- a/src/pkgcheck/checks/eclass.py
+++ b/src/pkgcheck/checks/eclass.py
@@ -61,6 +61,7 @@ class EclassUsageCheck(Check):
         'bash-completion': 'bash-completion-r1',
         'boost-utils': None,
         'clutter': 'gnome2',
+        'cmake-utils': 'cmake',
         'confutils': None,
         'darcs': None,
         'distutils': 'distutils-r1',


radhermit:
At this point I'm looking at removing the hardcoded deprecated eclass list as mentioned in https://github.com/pkgcore/pkgcheck/issues/243 so maintaining it would be done by editing eclass doc tags in the tree.

sbraz:
@radhermit so it needs to be tagged here? https://github.com/gentoo/gentoo/blob/master/eclass/user.eclass

radhermit:
> @radhermit so it needs to be tagged here? https://github.com/gentoo/gentoo/blob/master/eclass/user.eclass

The ``@DEPRECATED`` tag isn't added or documented yet for eclass docs in the devmanual. It's something I just made up since I think that'll be better going forward with the upcoming eclass doc parsing support in pkgcheck (https://github.com/pkgcore/pkgcheck/issues/74).

So for this case, you'd add a ``@DEPRECATED`` tag to the ``@ECLASS`` block in cmake-utils.eclass pointing to the cmake.eclass as its replacement. That would then get pulled and cached from the eclass during eclass doc parsing and then used to build up a more dynamic deprecated eclass list.

radhermit:
Closing since the hardcoded list was dropped in https://github.com/pkgcore/pkgcheck/commit/e6808185c46dfd1b33975870c8316cf3df184ad5. Updates for eclasses in the gentoo repo and devmanual info about the ``@DEPRECATED:`` tag will be completed before the next release.

PR #242, opened 2020-09-19
checks.repo_metadata: Detect moves to self

mgorny:
Detect meaningless move/slotmove entries that use the same package/slot
as both source and target.  This will hopefully catch sed mistakes.


radhermit:
Thanks, merged and changed the related result name to RedundantPackageUpdate, instead of MoveToSelfPackageUpdate.

PR #248, opened 2020-11-06
metadata_xml: Support grabbing metadata.xsd from the repo

mgorny:


radhermit:
This seems unnecessarily complex when all you probably want is a135eb74d08 reverted.

mgorny:
Technically speaking, the master-aware variant is more correct.

radhermit:
> Technically speaking, the master-aware variant is more correct.

Fair enough, merged with some changes on top. Thanks.

PR #306, opened 2021-03-14
PackageMetadataXmlCheck: refactor proxy-related checks

mgorny:
Refactor proxy-related check to be closer to the new GLEP 67 split.
The maintainers are grouped by type: proxied (yes), dev (no) and proxy.
The proxy-maint project check is generalized into a more generic proxy
without proxied maintainer check.  The support for explicit proxied=""
attribute will follow.


mgorny:
The second commit needs pkgcore/pkgcore#300.

radhermit:
Thanks, merged with minor wording and logic changes.

PR #322, opened 2021-05-09
python: Fix treating python3.10 as newer than python3.9

mgorny:
Fix the sorting logic to use a combined lexical-numerical sort, in order
to sort python3.10 as newer than python3.9.  This fixes
PythonCompatUpdate check.

### DIFF
--- a/src/pkgcheck/checks/python.py
+++ b/src/pkgcheck/checks/python.py
@@ -1,4 +1,5 @@
 import itertools
+import re
 
 from pkgcore.ebuild.atom import atom
 from pkgcore.restrictions import packages, values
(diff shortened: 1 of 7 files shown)


radhermit:
Wouldn't it make sense to add a USE flag sorting function to pkgcore that could be leveraged here? I guess it could be tricky to avoid special casing.

mgorny:
Maybe. I suppose a sorting function like this (i.e. combined numeric/non-numeric sort) could even be added to snakeoil.

radhermit:
Thanks, merged with minor changes in https://github.com/pkgcore/pkgcheck/commit/12a6ca0bb62766f7674644e91378b2ab035769bf.

More generic sorting functions can be looked into later.

radhermit:
> More generic sorting functions can be looked into later.

To follow up, pkgcore now provides functionality to generate sorting methods for any given USE_EXPAND group leveraging the flag ordering from the related profiles/desc/*.desc files. This should be more configurable, efficient, and simpler than trying to special-case a more generic sorting function.

mgorny:
Sounds like a good idea.

PR #333, opened 2021-07-16
PythonCompatUpdate: add overlays support

CyberTailor:
Parsing profiles/desc/python_targets.desc worked for ::gentoo only.

This commit changes previous logic to processing `python-exec2c` output,
which produces identical results but works everywhere.

(I'm unsure whether sorting is still needed)

### DIFF
--- a/src/pkgcheck/checks/python.py
+++ b/src/pkgcheck/checks/python.py
@@ -1,4 +1,5 @@
 import itertools
+import subprocess
 
 from pkgcore.ebuild.atom import atom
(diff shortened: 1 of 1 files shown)


radhermit:
I'd rather see the same code be used except forcibly targeted at the gentoo repo or whatever repo defines the flags... usually the default one with the lowest priority because separate processes should not be run during check initialization for a number of reasons.

radhermit:
Uh, I didn't mean to dissuade you into closing the PR, what I'm describing is probably easier than what you already did. :smile:

PR #334, opened 2021-07-16
PythonCheckCompat: add overlays support

CyberTailor:


mgorny:
Just to be clear (I'm new to overlay support in pkgcore), what order are `target_repos` in? `::gentoo` first or overlay first?

CyberTailor:
`self.options.target_repo` is an overlay
`self.options.target_repo.trees` is that overlay with its masters (masters go first)

src/pkgcheck/checks/metadata.py
```diff
@@ -733,11 +733,15 @@ class DependencyCheck(Check):
...
-        self.deprecated = self.options.target_repo.deprecated.match
         self.iuse_filter = use_addon.get_filter()
         self.conditional_ops = {'?', '='}
         self.use_defaults = {'(+)', '(-)'}
 
+        restrictions = set()
+        for repo in self.options.target_repo.trees:
+            restrictions.update(repo.deprecated.restrictions)
+        self.deprecated = boolean.OrRestriction(*restrictions, finalize=True).match
+
```
radhermit, line 744: I'd move this inherited support directly into pkgcore under ebuild repo objects. See https://github.com/pkgcore/pkgcore/commit/8b3b21b2426aa87ca7f9bd32bd66657dcb047cb3 as an example of what I'm talking about.

src/pkgcheck/checks/python.py
```diff
@@ -274,30 +274,34 @@ class PythonCompatCheck(Check):
...
+        # get Python implementations from profiles/desc/python_targets.desc
+        # or profiles/desc/python_single_target.desc
+        use_expand_desc = set()
+        for repo in self.options.target_repo.trees:
+            use_expand_desc.update(repo.config.use_expand_desc.get(prefix[:-1], ()))
+
+        for target, _desc in use_expand_desc:
+            if target[len(prefix):].startswith('python'):
+                yield target[len(prefix):]
+
```
radhermit, line 304: Honestly, I'd probably make this more generic and move it into pkgcore as well. Something like a function that can retrieve a given, inherited value from use_expand_desc across a target repo's dependents.

CyberTailor:
This is ready to be merged

PR #337, opened 2021-07-31
EapiCheck: detect using stable keywords on EAPI listed as testing

mgorny:


src/pkgcheck/checks/metadata.py
```diff
@@ -252,6 +252,19 @@ class BannedEapi(_EapiResult, results.Error):
     _type = 'banned'
 
 
+class StableKeywordsOnTestingEapi(results.VersionResult, results.Error):
+    """Package has stable keywords on EAPI marked as testing-only."""
+
+    def __init__(self, eapi, keywords, **kwargs):
+        super().__init__(**kwargs)
+        self.eapi = str(eapi)
+        self.keywords = sorted(keywords)
```
radhermit, line 261: Keyword class attributes should always be immutable. If you look at other examples, the sorted() call is usually done before the instance is created, and then tuple() is used for the keyword attribute.

src/pkgcheck/checks/metadata.py
```diff
@@ -282,6 +296,11 @@ def feed(self, pkg):
         elif eapi_str in self.options.target_repo.config.eapis_deprecated:
             yield DeprecatedEapi(pkg.eapi, pkg=pkg)
 
+        if eapi_str in self.options.target_repo.config.eapis_testing:
+            stable_keywords = [k for k in pkg.keywords if not k.startswith(('~', '-'))]
+            if stable_keywords:
+                yield StableKeywordsOnTestingEapi(pkg.eapi, stable_keywords, pkg=pkg)
```
radhermit, line 302: So in this case you'd do something like:

```python
if stable_keywords := sorted(k for k in pkg.keywords if not k.startswith(('~', '-'))):
```

and then

```python
self.keywords = tuple(keywords)
```

in the StableKeywordsOnTestingEapi class.

radhermit: I will also note that pkgcheck generally uses assignment expressions where it makes sense, so if you want to keep internal usage consistent you'd use one here.

mgorny: Wouldn't it exceed line width though?

radhermit: > Wouldn't it exceed line width though?

I didn't check, but pkgcheck et al. use a general loose standard of 100 chars so if exceeds that, I guess it's fine.

radhermit: Having checked, nope it's 98 chars when using the assignment op. :laughing:

radhermit: Anyway, I'd generally split it into a separate generator var, and then use the assignment operator along with sorted() which wouldn't even get close to 100 chars.

mgorny: I'm sorry, it's just been a long day.

mgorny:
I've finally gotten the tests to work! It's a miracle, valhalleluja!

PR #341, opened 2021-08-09
RepoProfilesCheck: use inheritance instead of lexical EAPI name cmp

mgorny:
Use profile inheritance to determine whether the 'newer' EAPI is
a superset of an 'older' EAPI rather than lexical comparison.

### DIFF
--- a/src/pkgcheck/checks/profiles.py
+++ b/src/pkgcheck/checks/profiles.py
@@ -411,9 +411,7 @@ class RepoProfilesCheck(RepoCheck):
                 continue
             for parent in profile.stack:
                 seen_profile_dirs.update(dir_parents(parent.name))
-                # flag lagging profile EAPIs -- assumes EAPIs are sequentially
-                # numbered which should be the case for the gentoo repo
-                if (self.options.gentoo_repo and str(profile.eapi) < str(parent.eapi)):
+                if profile.eapi is not parent.eapi and profile.eapi in parent.eapi.inherits:
                     lagging_profile_eapi[profile].append(parent)
 
         for profile, parents in lagging_profile_eapi.items():


mgorny:
CC @radhermit, @ulm

src/pkgcheck/checks/profiles.py
```diff
@@ -411,9 +411,8 @@ def finish(self):
                 continue
             for parent in profile.stack:
                 seen_profile_dirs.update(dir_parents(parent.name))
-                # flag lagging profile EAPIs -- assumes EAPIs are sequentially
-                # numbered which should be the case for the gentoo repo
-                if (self.options.gentoo_repo and str(profile.eapi) < str(parent.eapi)):
+                if (str(profile.eapi) != str(parent.eapi) and
```
radhermit, line 414: Why is string coercing needed here? It should be possible to compare EAPI objects directly for this whole change.

mgorny: I didn't trust the code ;-).

mgorny: Ok, changed now.

src/pkgcheck/checks/profiles.py
```diff
@@ -411,9 +411,7 @@ def finish(self):
                 continue
             for parent in profile.stack:
                 seen_profile_dirs.update(dir_parents(parent.name))
-                # flag lagging profile EAPIs -- assumes EAPIs are sequentially
-                # numbered which should be the case for the gentoo repo
-                if (self.options.gentoo_repo and str(profile.eapi) < str(parent.eapi)):
+                if profile.eapi != parent.eapi and profile.eapi in parent.eapi.inherits:
```
radhermit, line 414: I should have mentioned as well, I'd use the ``is`` operator rather than equality since EAPI objects are immutable, i.e.

```python
if profile.eapi is not parent.eapi
```

IIRC, I don't think the EAPI class even explicitly defines the related equality magic methods, but I guess they might come from auto-magic snakeoil inheritance in some fashion.

radhermit: Ah nevermind, I forgot that python uses the ``is`` operator internally if you don't provide a specific implementation for the __eq__, but personally I'd still consider using it explicitly instead.

mgorny: Sure.

PR #347, opened 2021-08-28
Add a check for eclasses calling EXPORT_FUNCTIONS before inherit

mgorny:
The behavior for handling EXPORT_FUNCTIONS before inherit is not
consistent across package managers.  Enforce calling it after inherit
to guarantee consistent behavior.


ulm:
Please put this on hold until discussion in QA and PMS teams (and on gentoo-dev) is finished.

PR #348, opened 2021-08-31
codingstyle: Gather indirect_eclasses recursively

mgorny:
Gather indirect_eclasses recursively to correctly handle eclass provide
chains, e.g. distutils-r1 → python-r1 → python-utils-r1.


mgorny:
Obsoleted by support in pkgcore.

PR #349, opened 2021-08-31
Add initial test for IndirectInherits test

mgorny:
Add an initial test verifying that IndirectInherits is reported
for the most basic test case.  Further tests will follow once @PROVIDES
support lands in pkgcore.


mgorny:
671650d984b47c0942a6fbdb6e5fa3387a149fd3

PR #357, opened 2021-12-23
setup.py: don't import setuptools._distutils directly

floppym:
Bug: https://bugs.gentoo.org/822537


mgorny:
Hmm, I guess I need to release snakeoil first.

PR #362, opened 2022-04-13
Add check for ebegin-eend calls being properly paired

laumann:
It was suggested to add QA warnings in Portage for calls to ebegin
missing a corresponding call to eend, and vice versa. Subsequently, it
was suggested that such lints belong in pkgcheck.

The idea is to use tree-sitter-bash to get a decent approximation for
calls to ebegin with an eend missing, and calls to eend where ebegin
hasn't been called.

Bug: https://bugs.gentoo.org/835823
Bug: https://bugs.gentoo.org/835824


laumann:
@radhermit @arthurzam This is obviously insufficient, but so far I've figured out how to add new checks, and the rudimentary use of tree-sitter-bash. I think I understand how to obtain and iterate over results from an s-expression, but I haven't yet grokked _how_ to write an s-expression to achieve what I want.

 - I think this is just an "Info" check (that's the lowest level right?), not a "Style"
 - this check can also be applied to eclasses, I think, but I don't know yet how to make that happen
 - write an s-expression and use it

laumann:
As for testing, I have a very simplistic ebuild that

<details>
<summary>dev-libs/foobar/foobar-0.0.1.ebuild</summary>

```
# Copyright 2022 Gentoo Authors
# Distributed under the terms of the GNU General Public License v2

EAPI=8

DESCRIPTION="Testing package"
HOMEPAGE="http://nowhere.com"

LICENSE="GPL-2"
SLOT="0"

pkg_pretend() {
        ebegin "well hello there"
}

pkg_setup() {
        eend 1
}
```
</details>

with which I get this output:

```
repo
  EbeginMissingEend: Call to ebegin with a missing call to eend (pkg_pretend)
  EendMissingEbegin: Call to eend, but missing preceding call to ebegin (pkg_setup)
```
How do I get the package name (dev-libs/foobar) instead of "repo"?

laumann:
Oustanding items:
 - [ ] Handle control flow/scoping
 - [ ] Testing

src/pkgcheck/checks/codingstyle.py
```diff
@@ -79,6 +79,64 @@ def feed(self, pkg):
                     yield DeprecatedEapiCommand(name, line=call, lineno=lineno+1, eapi=pkg.eapi, pkg=pkg)
 
 
+class EbeginMissingEend(results.LineResult, results.Style):
```
radhermit, line 82: This should be a warning result, not a stylistic one.

src/pkgcheck/checks/codingstyle.py
```diff
@@ -79,6 +79,64 @@ def feed(self, pkg):
...
+    """Ebuild calls ebegin in a phase, but is missing a call to eend"""
+
+    def __init__(self, func_name, **kwargs):
+        super().__init__(**kwargs)
+        self.func_name = func_name
+
+    @property
+    def desc(self):
+        return f'Call to ebegin with a missing call to eend ({self.func_name}), on line {self.lineno}'
```
radhermit, line 91: To match the brevity and formatting of result descriptions this should be something like:

```python
return f'ebegin missing corresponding eend, line {self.lineno}'
```

src/pkgcheck/checks/codingstyle.py
```diff
@@ -79,6 +79,64 @@ def feed(self, pkg):
...
+class EendMissingEbegin(results.LineResult, results.Style):
+    """Ebuild calls eend in a phase, but is missing a call to ebegin"""
+
+    def __init__(self, func_name, **kwargs):
+        super().__init__(**kwargs)
+        self.func_name = func_name
+
+    @property
+    def desc(self):
+        return f'Call to eend, but missing preceding call to ebegin ({self.func_name}), on line {self.lineno}'
```
radhermit, line 103: Simlar to the above:

```python
return f'eend missing corresponding ebegin, line {self.lineno}'
```

src/pkgcheck/checks/codingstyle.py
```diff
@@ -79,6 +79,64 @@ def feed(self, pkg):
...
+
+
+class EbeginEendCheck(Check):
+    """Scan ebuild for ebegin with missing eend, and vice versa"""
+    #probably also should try to see if you can handle scoping at all, or at least use a stack design to handle multiple calls
+
+    _source = sources.EbuildParseRepoSource
+    known_results = frozenset([EbeginMissingEend, EendMissingEbegin])
```
radhermit, line 111: I'd combine both of these results into one (e.g. UnmatchedEbeginEend) using a stack design or similar to find them in the check like I mentioned before.

src/pkgcheck/checks/codingstyle.py
```diff
@@ -79,6 +79,64 @@ def feed(self, pkg):
                     yield DeprecatedEapiCommand(name, line=call, lineno=lineno+1, eapi=pkg.eapi, pkg=pkg)
 
 
+class EbeginMissingEend(results.LineResult, results.Style):
+    """Ebuild calls ebegin in a phase, but is missing a call to eend"""
```
radhermit, line 83: Docstrings are generally sentences in pkgcheck, so use a period.

src/pkgcheck/checks/codingstyle.py
laumann: I'm working on it, going for the stack design, but it's proving a bit of a challenge for me :-)

laumann:
I've given up on this one (for now). I much prefer the little dynamic checks that exist in portage today than trying to implement a static analysis check that I fear will continuously be extended with edge cases.

So I'll just close this :-)

PR #364, opened 2022-04-19
pkgcheck/sources: properly close resources

laumann:
Testing with PYTHONWARNINGS=d,i::ImportWarning pointed out a few places
where allocated resources were being leaked.


mgorny (approved):
Nice catch! Thanks!

PR #365, opened 2022-04-20
new check: calls to eend without an argument

laumann:
Calls to eend should always have an argument.

### DIFF
--- a/src/pkgcheck/checks/codingstyle.py
+++ b/src/pkgcheck/checks/codingstyle.py
@@ -79,6 +79,29 @@ class BadCommandsCheck(Check):
                     yield DeprecatedEapiCommand(name, line=call, lineno=lineno+1, eapi=pkg.eapi, pkg=pkg)
 
 
+class EendMissingArg(results.LineResult, results.Warning):
+    """Ebuild calls eend with no arguments."""
+
+    @property
+    def desc(self):
+        return f'eend with no arguments, on line {self.lineno}'
+
+
+class EendMissingArgCheck(Check):
+    """Scan an ebuild for calls to eend with no arguments."""
+
+    _source = sources.EbuildParseRepoSource
+    known_results = frozenset([EendMissingArg])
+
+    def feed(self, pkg):
+        for func_node, _ in bash.func_query.captures(pkg.tree.root_node):
+            for node, _ in bash.cmd_query.captures(func_node):
+                line = pkg.node_str(node)
+                if line == "eend":
+                    lineno, _ = node.start_point
+                    yield EendMissingArg(line=line, lineno=lineno+1, pkg=pkg)
+
+
 class MissingSlash(results.VersionResult, results.Error):
     """Ebuild uses a path variable missing a trailing slash."""
 
--- a/testdata/data/repos/standalone/EendMissingArgCheck/EendMissingArg/expected.json
+++ b/testdata/data/repos/standalone/EendMissingArgCheck/EendMissingArg/expected.json
@@ -1,0 +1,1 @@
+{"__class__": "EendMissingArg", "category": "EendMissingArgCheck", "package": "EendMissingArg", "version": "0", "line": "eend", "lineno": 10}
--- a/testdata/data/repos/standalone/EendMissingArgCheck/EendMissingArg/fix.patch
+++ b/testdata/data/repos/standalone/EendMissingArgCheck/EendMissingArg/fix.patch
@@ -1,0 +1,9 @@
+--- standalone/EendMissingArgCheck/EendMissingArg/EendMissingArg-0.ebuild	2022-04-20 14:25:14.385255909 +0200
++++ fixed/EendMissingArgCheck/EendMissingArg/EendMissingArg-0.ebuild	2022-04-20 15:42:04.363051160 +0200
+@@ -7,5 +7,5 @@
+ 
+ src_install() {
+ 	ebegin "installing"
+-	eend
++	eend $?
+ }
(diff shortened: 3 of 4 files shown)


src/pkgcheck/checks/codingstyle.py
```diff
@@ -79,6 +79,29 @@ def feed(self, pkg):
...
+    """Scan an ebuild for calls to eend without an argument"""
+
+    _source = sources.EbuildParseRepoSource
+    known_results = frozenset([EendMissingArg])
+
+    def feed(self, pkg):
+        for func_node, _ in bash.func_query.captures(pkg.tree.root_node):
+            for node, _ in bash.cmd_query.captures(func_node):
+                call = pkg.node_str(node)
+                if call == "eend":
```
laumann, line 100: I'm sure with the right s-expression, this could be reduced to one query, iterating over the result and reporting each hit.

src/pkgcheck/checks/codingstyle.py
```diff
@@ -79,6 +79,29 @@ def feed(self, pkg):
...
 
+class EendMissingArg(results.LineResult, results.Style):
+    """Eend is called without an argument"""
+
+    @property
+    def desc(self):
+        return f'Call to eend is missing an argument, on line {self.lineno}'
+
+
+class EendMissingArgCheck(Check):
```
radhermit, line 90: I'd combine this check with the previous ebegin/eend related check.

src/pkgcheck/checks/codingstyle.py
```diff
@@ -79,6 +79,29 @@ def feed(self, pkg):
                     yield DeprecatedEapiCommand(name, line=call, lineno=lineno+1, eapi=pkg.eapi, pkg=pkg)
 
 
+class EendMissingArg(results.LineResult, results.Style):
+    """Eend is called without an argument"""
+
+    @property
+    def desc(self):
+        return f'Call to eend is missing an argument, on line {self.lineno}'
```
radhermit, line 87: Reformat this similar to how I showed in the previous PR.

src/pkgcheck/checks/codingstyle.py
```diff
@@ -79,6 +79,29 @@ def feed(self, pkg):
                     yield DeprecatedEapiCommand(name, line=call, lineno=lineno+1, eapi=pkg.eapi, pkg=pkg)
 
 
+class EendMissingArg(results.LineResult, results.Style):
+    """Eend is called without an argument"""
```
radhermit, line 83: Docstring sentence formatting same as the previous PR.

src/pkgcheck/checks/codingstyle.py
```diff
@@ -79,6 +79,29 @@ def feed(self, pkg):
                     yield DeprecatedEapiCommand(name, line=call, lineno=lineno+1, eapi=pkg.eapi, pkg=pkg)
 
 
+class EendMissingArg(results.LineResult, results.Style):
```
radhermit, line 82: Should be a warning result, not style.

src/pkgcheck/checks/codingstyle.py
laumann: I'd rather not do that at this moment. This check is simple, because it can just focus on one line, the ebegin-eend matching requires keeping some context around.

If they're combined, I'd rather have this added first, and then return with an implementation for ebegin-eend.

src/pkgcheck/checks/codingstyle.py
laumann: I'm not clear on the distinction here. `eend` doesn't fail, but doesn't do much either (other than print `[ok]`). But I'll change it.

thesamesam: It's warning here given it's illegal, from PMS:
>    Indicates that the process begun with an ebegin message has completed. Takes one fixed argument, which is a   numeric return code, and an optional message in all subsequent arguments. If the first argument is 0, prints a success indicator; otherwise, prints the message followed by a failure indicator. Returns its first argument as exit status.

The idea here being: it's not just poor taste or a bit dodgy, it's definitely _wrong_. Portage shouldn't even accept it and won't in the future.

laumann: Oh, I wasn't aware that it's actually required. Makes way more sense as a warning then, thanks @thesamesam

radhermit: Personally I'd advise avoiding using portage as a standard for what is acceptable in many cases and look directly at the spec instead. :laughing:

src/pkgcheck/checks/codingstyle.py
laumann: Done.

src/pkgcheck/checks/codingstyle.py
laumann: Done.

src/pkgcheck/checks/codingstyle.py
laumann: > Personally I'd advise avoiding using portage as a standard for what is acceptable 

Good point :smile:

laumann: Well, once https://github.com/gentoo/gentoo/pull/25134 is merged, these should all be fixed, and portage can make it a hard error :tada:

laumann: ... or maybe not, that might break a bunch of overlays.

thesamesam: That's not a huge problem 😉

radhermit: It just reinforces my viewpoint that all these kind of checks should be also done in a linting tool so overlay maintainers can easily scan and fix their issues without having to build every package in their tree while hoping the situation isn't conditionally triggered.

thesamesam:
@arthurzam Is this one OK to merge now? I think @laumann wanted to do any remaining stuff as a follow-up if it goes OK.

arthurzam (approved):
Looks very good - thank you @laumann

PR #367, opened 2022-04-24
bash: use libtree-sitter-bash.so from system if available

laumann:
Prefer to load a shared library already installed for tree-sitter-bash.
Tested on Gentoo by emerging dev-libs/tree-sitter-bash.


laumann:
A few things I thought about

 - should it be a user-configurable thing to choose which tree-sitter-bash version to use?
 - should it limit this to specific systems? (as @radhermit pointed out: the issues you may have debundling it are broken OS X support and broken pip builds)
 - it really should check that the version of tree-sitter-bash on the system is >=0.19.0 (the current version bundled)

src/pkgcheck/bash/__init__.py
```diff
@@ -92,8 +94,10 @@ def build_library(output_path, repo_paths):  # pragma: no cover
         build_library(lib, [bash_src])
 
 
-if os.path.exists(lib):
-    lang = Language(lib, 'bash')
+syslib = find_library('tree-sitter-bash')
+
+if syslib or os.path.exists(lib):
+    lang = Language(syslib if syslib else lib, 'bash')
```
mgorny, line 100: I think

```suggestion
    lang = Language(syslib or lib, 'bash')
```

will work too.

laumann: Neat! Thanks.

laumann:
Hey, there's even a bug for it: https://bugs.gentoo.org/834135

src/pkgcheck/bash/__init__.py
```diff
@@ -87,13 +91,12 @@ def build_library(output_path, repo_paths):  # pragma: no cover
     from .. import _const
 except ImportError:  # pragma: no cover
     # build library when running from git repo or tarball
-    if not os.path.exists(lib) and 'tree-sitter-bash' in os.listdir(const.REPO_PATH):
+    if not syslib and not os.path.exists(lib) and 'tree-sitter-bash' in os.listdir(const.REPO_PATH):
```
laumann, line 94: As I understood this logic, the `_const.py` module is generated on build/installation (at least on Gentoo), and, as part of that, the vendored tree-sitter-bash is built (as `lang.so`). When running from checkout the _const module doesn't exist and so this ImportError is always thrown (and the vendored lib is built).

mgorny, line 94:
```suggestion
    if syslib is None and not os.path.exists(lib) and 'tree-sitter-bash' in os.listdir(const.REPO_PATH):
```

I think in general we prefer explicitly checking for `None` even no other valid value would evaluate false.

laumann:
@arthurzam @radhermit any feedback on this?

arthurzam (approved):

laumann:
Document tree-sitter-bash as a dependency.

mgorny:
Could you update the CI test matrix to test both building against system and bundled tree-sitter-bash?

laumann:
> Could you update the CI test matrix to test both building against system and bundled tree-sitter-bash?

Sure.

MatthewGentoo:
> it really should check that the version of tree-sitter-bash on the system is >=0.19.0 (the current version bundled). EDIT: The language version is 13, the release is v0.19.0. The language version is available in the TSLanguage struct, so all that's really needed is to fish it out (I think).

The language version (13) determines which versions of tree-sitter the grammar is compatible with. I don't think we need to worry about it too much because newer tree-sitter versions are backwards compatible with old grammars.

I think the problem is if a new version renames or removes any node types because this would break pkgcheck's queries. Unsure on what upstream's policy for that is, but if you look at [the commit history for the grammar](https://github.com/tree-sitter/tree-sitter-bash/commits/master/src/grammar.json) it seems stable.

laumann:
> The language version (13) determines which versions of tree-sitter the grammar is compatible with. I don't think we need to worry about it too much because newer tree-sitter versions are backwards compatible with old grammars.

OK, I won't worry about it then, thanks :-)

laumann:
> Could you update the CI test matrix to test both building against system and bundled tree-sitter-bash?

@mgorny I started looking at this, but it turns out that it'll require fetching a snapshot of tree-sitter-bash, building and manually installing it to get it working as ubuntu doesn't package libtree-sitter-bash yet (only libtree-sitter-dev exists afaict).

I can do it (I think), but it's gonna take a bit more work than I anticipated.

.github/workflows/test.yml
```diff
@@ -15,6 +15,7 @@ jobs:
         os: [ubuntu-latest, macos-latest]
         python-version: ['3.8', '3.9']
         deps: [newest-deps]
+        tree-sitter-bash: [system, bundled]
```
laumann, line 18: Better suggestions for names are welcome :-) the value "system" doesn't really mean much when looking at the build overview. A key like "tree-sitter-bash-sys" or something would probably be better.

.github/workflows/test.yml
```diff
@@ -69,6 +70,22 @@ jobs:
...
+    - name: Install tree-sitter-bash
+      if: ${{ matrix.os == 'ubuntu-latest' && matrix.tree-sitter-bash == 'system' }}
+      run: |
+        sudo apt install build-essential wget
+        wget https://github.com/tree-sitter/tree-sitter-bash/archive/v0.19.0.tar.gz
+        tar xvzf v0.19.0.tar.gz
+        pushd tree-sitter-bash-0.19.0/src
+        objects=( parser.o scanner.o )
+        make CFLAGS="-I${PWD}" CXXFLAGS="-I${PWD}" "${objects[@]}"
+        g++ -Wl,-O1 -Wl,--as-needed -shared parser.o scanner.o -Wl,--soname=libtree-sitter-bash.so.13 -o libtree-sitter-bash.so.13
```
laumann, line 84: This is heavily inspired by [tree-sitter-grammar.eclass](https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/tree-sitter-grammar.eclass#n67).

laumann:
I think this is as far as I'll take it for now. @mgorny has suggested that setup.py be replaced, maybe by flit, but I'd prefer to do that in another PR if that's OK.

Just to summarize:
 - bundled tree-sitter-bash is only built if bookl(USE_SYSTEM_TREE_SITTER_BASH) env var is true (default is False)
 - when running from checkout, the dynamic building only happens if the system lib isn't detected
 - CI is expanded to compile and "install" (ie set LD_LIBRARY_PATH) tree-sitter-bash - notably doesn't require libtree-sitter-dev ubuntu package. Only tested on ubuntu
 - document tree-sitter-bash as a dependency in README

.github/workflows/test.yml
```diff
@@ -69,6 +75,23 @@ jobs:
...
         mv requirements/install.txt requirements/dev.txt
 
+    # To use libtree-sitter-bash.so installed on the system, we need to compile
+    # it manually, because ubuntu doesn't yet package (but Gentoo does)
+    - name: Install tree-sitter-bash
+      if: ${{ matrix.os == 'ubuntu-latest' && matrix.tree-sitter-bash == 'system' }}
+      run: |
+        sudo apt install build-essential wget
+        wget https://github.com/tree-sitter/tree-sitter-bash/archive/v0.19.0.tar.gz
+        tar xvzf v0.19.0.tar.gz
```
mgorny, line 85:
```suggestion
        tar -xvzf v0.19.0.tar.gz
```

Let's not use the ancient syntax ;-).

.github/workflows/test.yml
```diff
@@ -83,6 +106,23 @@ jobs:
...
+        fi
+
+    - name: Assert that bundled version is built
+      if: ${{ matrix.os == 'ubuntu-latest' && matrix.tree-sitter-bash == 'bundled' }}
+      run: |
+        if [[ ! -n "$(find -name 'lang.so')" ]]; then
+          echo "::error Bundled tree-sitter-bash was not built!"
+          false
+        fi
+
```
mgorny, line 125: How about merging these two into one, something like (untested):

```suggestion
    - name: Check whether bundled version was built
      if: ${{ matrix.os == 'ubuntu-latest' && matrix.tree-sitter-bash == 'system' }}
      run: |
        EXPECTED=${{ matrix.tree-sitter-bash == 'bundled' }}  # assuming it gets converted to 0/1
        if [[ "$(find -name 'lang.so' | wc -l)" == ${EXPECTED} ]]; then
          find -name 'lang.so'
          echo "::error Expected ${{ matrix.tree-sitter-bash }} tree-sitter, found the other"
          false
        fi

```

That said, I'm not sure if expressions can be used in commands, you may need to push them via envvar instead. Also bools could be true/false rather than 1/0.

mgorny (approved):
LGTM. But let's wait some in case @radhermit has any comments.

arthurzam (approved):
Thank you @laumann

.github/workflows/test.yml
laumann: Done.

thesamesam:
@laumann You can do that PR for 9999 now :wink:

Issue #366, opened 2022-04-24
New check: LICENSE doesn't contain variables


laumann:
To check for the [new QA policy](https://archives.gentoo.org/gentoo-dev/message/b7764a5112270858ea450ccfa508dc17).

> The only exception [...] is the use of LICENSE itself, i.e. appending is allowed

radhermit:
This should be a one-line change (not including tests).

laumann:
Thanks for the tips @radhermit I'll see about implementing this soon.

PR #368, opened 2022-04-30
new check: LICENSE should not contain variables

laumann:
Extend MetadataVarCheck with a new method checking the LICENSE variable
for other variables. The only exception is LICENSE itself. The check
accepts occurrences of $LICENSE and ${LICENSE}.

Closes: https://github.com/pkgcore/pkgcheck/issues/366

### DIFF
--- a/testdata/repos/standalone/MetadataVarCheck/ReferenceInMetadataVar/ReferenceInMetadataVar-2.ebuild
+++ b/testdata/repos/standalone/MetadataVarCheck/ReferenceInMetadataVar/ReferenceInMetadataVar-2.ebuild
@@ -1,0 +1,7 @@
+DESCRIPTION="Ebuild with variable in LICENSE"
+HOMEPAGE="https://github.com/pkgcore/pkgcheck"
+SLOT="0"
+MY_LICENSE="BSD"
+LICENSE="${MY_LICENSE}"
+LICENSE="${LICENSE} BSD"
+KEYWORDS="~amd64 ~x86"
--- a/src/pkgcheck/checks/codingstyle.py
+++ b/src/pkgcheck/checks/codingstyle.py
@@ -417,8 +417,13 @@ class ReferenceInMetadataVar(results.VersionResult, results.Style):
     KEYWORDS must be a simple string with literal content as stated by the QA
     policy guide [#]_.
 
+    LICENSE must specify all license names verbatim, without referring to any
+    variables. The only exception is the LICENSE variable itself, ie appending
+    is allowed [#]_.
+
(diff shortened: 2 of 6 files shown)


thesamesam (approved):

testdata/repos/standalone/MetadataVarCheck/ReferenceInMetadataVar/ReferenceInMetadataVar-2.ebuild
```diff
@@ -0,0 +1,6 @@
+DESCRIPTION="Ebuild with variable in LICENSE"
```
mgorny, line 1: Hmm, maybe it'd also worth testing that the check works over `+=` in addition to `=`.

```
LICENSE="foo"  # ok
LICENSE+=" ${bar}"  # not ok
```

arthurzam:
@laumann 
Also, as our discussion at `#gentoo-qa`, please also add check for `LICENSE="${LICENSE} ${LICENSE/M/G}-2"`, so we know we disallow it.
In case we decide to allow it in future, we can test the behavior change.

laumann:
Hmm, seems i broke some other tests.

> please also add check for `LICENSE="${LICENSE} ${LICENSE/M/G}-2"`, so we know we disallow it.

I'll add it. I really doubt that it would be allowed though :-)

testdata/repos/standalone/MetadataVarCheck/ReferenceInMetadataVar/ReferenceInMetadataVar-2.ebuild
laumann: done.

mgorny (approved):
Thanks for doing this!

Could you also paste the current matches as a comment on this PR?

laumann:
> Could you also paste the current matches as a comment on this PR?

There aren't any! I think @ulm fixed them not too long ago.

laumann:
but if I check out for example d07e7ff26348a277be0d4d18fc63f6bdfea94038 I get the following hits:

```
dev-lang/mmix
  ReferenceInMetadataVar: version 20160804-r1: LICENSE includes variable: ${PN}

games-emulation/dolphin
  ReferenceInMetadataVar: version 5.0_p20210506-r3: LICENSE includes variable: KEEP_BUNDLED[*]
  ReferenceInMetadataVar: version 9999: LICENSE includes variable: KEEP_BUNDLED[*]

games-rpg/comi
  ReferenceInMetadataVar: version 1: LICENSE includes variable: ${PN}
```

arthurzam (approved):
Asking for last blocker from QA team or other devs before merging it.

Issue #374, opened 2022-05-13
Add DeprecatedEclassFunction


thesamesam:
Extremely similar to https://github.com/pkgcore/pkgcheck/issues/373.

pkgcheck should warn about usage of eclass functions which are marked as `@DEPRECATED` within eclassdocs.

For example, `go-module.eclass` marks `go-module_set_globals` as deprecated:
```
# @FUNCTION: go-module_set_globals
# @DEPRECATED: none
# @DESCRIPTION:
# Convert the information in EGO_SUM for other usage in the ebuild.
# - Populates EGO_SUM_SRC_URI that can be added to SRC_URI
# - Exports _GOMODULE_GOSUM_REVERSE_MAP which provides reverse mapping from
#   distfile back to the relative part of SRC_URI, as needed for
#   GOPROXY=file:///...
go-module_set_globals() {
```

Example bad ebuild:
```
$ cat app-admin/gopass/gopass-1.13.0.ebuild
# Copyright 1999-2021 Gentoo Authors
# Distributed under the terms of the GNU General Public License v2

EAPI=8

inherit go-module optfeature

EGO_SUM=(
        "cloud.google.com/go v0.26.0/go.mod"
        "cloud.google.com/go v0.34.0/go.mod"
        "cloud.google.com/go v0.38.0/go.mod"
        "cloud.google.com/go v0.44.1/go.mod"
[...]
        "rsc.io/qr v0.2.0/go.mod"
        "rsc.io/quote/v3 v3.1.0/go.mod"
        "rsc.io/sampler v1.3.0/go.mod"
)
go-module_set_globals

SRC_URI=...
[...]
```

pkgcheck should warn about the usage of `go-module_set_globals` and say it is "deprecated (replacement: none)" in this case.

Issue #373, opened 2022-05-13
Add DeprecatedEclassVariable


thesamesam:
This would be useful to catch deprecations of techniques from eclasses, e.g.:
- `DISTUTILS_USE_SETUPTOOLS` from `distutils-r1.eclass` (replaced by `DISTUTILS_USE_PEP517`)
- `EGO_SUM` from `go-module.eclass` (no replacement, use dep tarballs)

We already have `@DEPRECATED` tags for eclass variables which allow optionally specifying a replacement (or "none").

Example eclass marking something as deprecated (this is from `go-module.eclass`):
```
# @ECLASS_VARIABLE: EGO_SUM
# @DEPRECATED: none
# @DESCRIPTION:
# This is replaced by a dependency tarball, see above for how to create
# one.
```

Example bad ebuild:
```
$ cat app-admin/gopass/gopass-1.13.0.ebuild
# Copyright 1999-2021 Gentoo Authors
# Distributed under the terms of the GNU General Public License v2

EAPI=8

inherit go-module optfeature

EGO_SUM=(
        "cloud.google.com/go v0.26.0/go.mod"
        "cloud.google.com/go v0.34.0/go.mod"
        "cloud.google.com/go v0.38.0/go.mod"
        "cloud.google.com/go v0.44.1/go.mod"
[...]
```

pkgcheck should warn about the usage of `EGO_SUM` and say it is "deprecated (replacement: none)" in this case.

PR #375, opened 2022-05-13
new check: usage of deprecated variables or functions

arthurzam:
Resolves: https://github.com/pkgcore/pkgcheck/issues/373
Resolves: https://github.com/pkgcore/pkgcheck/issues/374

<details>
<summary>Example output</summary>

```
app-admin/gopass
  UnstableOnly: for arches: [ ppc64, x86 ], all versions are unstable: [ 1.12.6, 1.12.7, 1.13.0 ]
  DeprecatedEclassVariable: version 1.12.6: uses deprecated variable on line 8: EGO_SUM (no replacement)
  DeprecatedEclassFunction: version 1.12.6: uses deprecated function on line 494: go-module_set_globals (no replacement)
  PotentialStable: version 1.12.6: slot(0), stabled arch: [ amd64 ], potentials: [ ~ppc64, ~x86 ]
  DeprecatedEclassVariable: version 1.12.7: uses deprecated variable on line 8: EGO_SUM (no replacement)
  DeprecatedEclassFunction: version 1.12.7: uses deprecated function on line 498: go-module_set_globals (no replacement)
  RedundantVersion: version 1.12.7: slot(0) keywords are overshadowed by version: 1.13.0
  DeprecatedEclassVariable: version 1.13.0: uses deprecated variable on line 8: EGO_SUM (no replacement)
  DeprecatedEclassFunction: version 1.13.0: uses deprecated function on line 510: go-module_set_globals (no replacement)
  StableRequest: version 1.13.0: slot(0) no change in 146 days for unstable keyword: [ ~amd64 ]
```
</details>

---

TODO:

- [x] Add tests for this new check

### DIFF
--- a/src/pkgcheck/checks/eclass.py
+++ b/src/pkgcheck/checks/eclass.py
@@ -30,6 +30,40 @@ class DeprecatedEclass(results.VersionResult, results.Warning):
         return f'uses deprecated eclass: {self.eclass} ({replacement})'
 
 
+class DeprecatedEclassVariable(results.LineResult, results.Warning):
+    """Package uses a deprecated variable from an eclass."""
+
+    def __init__(self, variable, replacement, **kwargs):
+        super().__init__(**kwargs)
+        self.variable = variable
+        self.replacement = replacement
+
+    @property
+    def desc(self):
+        if self.replacement is not None:
+            replacement = f'migrate to {self.replacement}'
+        else:
+            replacement = 'no replacement'
+        return f'uses deprecated variable on line {self.lineno}: {self.variable} ({replacement})'
+
+
+class DeprecatedEclassFunction(results.LineResult, results.Warning):
+    """Package uses a deprecated function from an eclass."""
+
+    def __init__(self, function, replacement, **kwargs):
+        super().__init__(**kwargs)
+        self.function = function
+        self.replacement = replacement
+
+    @property
+    def desc(self):
+        if self.replacement is not None:
+            replacement = f'migrate to {self.replacement}'
+        else:
+            replacement = 'no replacement'
+        return f'uses deprecated function on line {self.lineno}: {self.function} ({replacement})'
+
+
 class DuplicateEclassInherit(results.LineResult, results.Style):
     """An ebuild directly inherits the same eclass multiple times.
 
@@ -69,7 +103,8 @@ class EclassUsageCheck(Check):
 
     _source = sources.EbuildParseRepoSource
     known_results = frozenset([
-        DeprecatedEclass, DuplicateEclassInherit, MisplacedEclassVar,
+        DeprecatedEclass, DeprecatedEclassVariable, DeprecatedEclassFunction,
+        DuplicateEclassInherit, MisplacedEclassVar,
     ])
     required_addons = (addons.eclass.EclassAddon,)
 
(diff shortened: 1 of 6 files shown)


src/pkgcheck/checks/eclass.py
```diff
@@ -30,6 +30,40 @@ def desc(self):
...
+    """Package uses a deprecated variable from an eclass."""
+
+    def __init__(self, variable, replacement, **kwargs):
+        super().__init__(**kwargs)
+        self.variable = variable
+        self.replacement = replacement
+
+    @property
+    def desc(self):
+        if self.replacement is not True:
```
thesamesam, line 43: We're using a different pattern here than above (`is not None` vs `is not True`). Why not just `if self.replacement` always?

thesamesam (commented):
Nice. Thanks for doing this so quickly. It seems reasonable to do the functions + variables together.

src/pkgcheck/checks/eclass.py
arthurzam: We have 3 states:

1. `None` - not deprecated
2. `True` - no replacement string
3. other - the replacement string

I'm not sure what is the tag for eclass and why was it correct in that check. If I use `is not None` in my code, I get
```
DeprecatedEclassFunction: version 1.13.0: uses deprecated function: go-module_set_globals (migrate to True)
```

thesamesam: Huh. OK, whatever. Thanks!

radhermit: A better contextual explanation (off the top of my head but probably still relevant): pkgcore eclassdoc parsing provides three states:

1. False - not deprecated (the default value)
2. True - deprecated with no replacement (as specified in the eclass if the replacement value is 'None' or 'none')
3. string value - deprecated with a given value as the replacement

pkgcheck maps this into two states for the deprecated eclass result (as it ignores the non-deprecated state):

1. None - deprecated with no replacement
2. string value - deprecated with a given value as a replacement

What should be done for all related results is what sam said (and probably was just overlooked during writing/refactoring eclassdoc parsing initially). To do that for this check just copy what is done for eclasses and map non-string values (True) to None for the replacement argument.

src/pkgcheck/checks/eclass.py
```diff
@@ -30,6 +30,40 @@ def desc(self):
...
+            replacement = f'migrate to {self.replacement}'
+        else:
+            replacement = 'no replacement'
+        return f'uses deprecated variable: {self.variable} ({replacement})'
+
+
+class DeprecatedEclassFunction(results.VersionResult, results.Warning):
```
radhermit, line 50: This should be a LineResult for more context for the user.

src/pkgcheck/checks/eclass.py
```diff
@@ -30,6 +30,40 @@ def desc(self):
         return f'uses deprecated eclass: {self.eclass} ({replacement})'
 
 
+class DeprecatedEclassVariable(results.VersionResult, results.Warning):
```
radhermit, line 33: This should be a LineResult for more context for the user.

src/pkgcheck/checks/eclass.py
```diff
@@ -98,6 +133,38 @@ def check_pre_inherits(self, pkg, inherits):
...
+                for var in self.eclass_cache[eclass].variables:
+                    if var.deprecated:
+                        deprecated_variables[var.name] = var.deprecated
+                for func in self.eclass_cache[eclass].functions:
+                    if func.deprecated:
+                        deprecated_functions[func.name] = func.deprecated
+
+        # scan for usage of @DEPRECATED variables
+        if deprecated_variables:
+            for node, _ in bash.var_assign_query.captures(pkg.tree.root_node):
```
radhermit, line 152: I'd probably use `var_query` here instead of `var_assign_query` as I imagine some deprecated, eclass variables can be used without assigning to them (assuming this doesn't add false positives for whatever reason).

arthurzam:
@radhermit Thanks for the advises and suggestions, it is indeed better that way. I checked and didn't find false positives (at ::gentoo tree we don't have a lot of deprecated functions or variables).
@thesamesam I have split the check function into two separate functions, as there wasn't a lot gain from one, and now it is easier to read.

I have also added tests (taken some from the real `go-module.eclass` and did some manual changes to better check deprecated text)

thesamesam (approved):
Great work and thanks for doing this so quickly too.

PR #383, opened 2022-05-23
new check: GH archives missing .gh suffix when pypi is available

### LINKED ISSUE #382: Add a check for packages having PyPI remote-id & missing .gh suffix for GitHub archives
I'd use something that detects the combination of following:

1. The package has a pypi remote-id.
2. the package fetches a GitHub archive (i.e. `^https://github.com/.*/archive/`, I guess)
3. the archive filename doesn't end with `.gh.tar.gz` suffix.

The rough idea is that we don't want conflicts between sdists published on pypi and GitHub archives.

mgorny:
Add a new check that verifies that Python packages using GitHub archives
use the `.gh.tar.gz` suffix to make them distinct from the sdist
tarballs on PyPI.

Fixes #382


thesamesam:
I don't really see the need to enforce .gh in tarball names given that it's somewhat common sense to need to rename it if you change for the same version? This is how we've handled it for quite some time.

arthurzam (changes requested):
Please add tests for:

- package with pypi remote-id, but without any github distfile
- for perfect coverage, we also need a package with pypi remote-id but without any SRC_URI. Otherwise (as this is niche), we can add skip coverage for the `for uri in f.uri` line.

Also, the same comment as previous PR about the commit message, please use
`Fixes: https://github.com/pkgcore/pkgcheck/issues/382`

arthurzam:
> I don't really see the need to enforce .gh in tarball names given that it's somewhat common sense to need to rename it if you change for the same version? This is how we've handled it for quite some time.

Maybe because I'm from "mgorny's python academy", I prefer to add the `.gh` to all python packages that use github tarballs. More clear, safer, and easier to read.

> I think this change makes more sense if the motivation is to just have clarity in distfiles names. But it doesn't really have much utility in preventing the "I changed from pypi to GitHub in the same release" case, as this is pretty rare.

Rare, but not impossible, and quite hard to catch and find, and I nearly done it, and even worse, was caught by mgorny in one PR (during the month before becoming a gentoo-dev).

thesamesam:
I'm not against it but I'd like the rationale to be documented at least within the commit message, and for something to be added to the Python Guide if it's not there already. I'd certainly missed it being added.

mgorny:
> I don't really see the need to enforce .gh in tarball names given that it's somewhat common sense to need to rename it if you change for the same version? This is how we've handled it for quite some time.
> 
> I think this change makes more sense if the motivation is to just have clarity in distfiles names. But it doesn't really have much utility in preventing the "I changed from pypi to GitHub in the same release" case, as this is pretty rare.

Sure, whatever works for you. I meant more "switching from GitHub to pypi, or from GitHub to both" (e.g. when you need some generated file). But yes, "following mgorny's style" is a good argument for me too ;-).

mgorny:
And yes, I'm pretty sure that at least once I've caught myself replacing SRC_URI and forgetting that I need to rename the file.

arthurzam (approved):

PR #384, opened 2022-05-23
network: Do not attempt to verify URL with unknown protocol

### LINKED ISSUE #372: Duplicate reports for unsupported protocol
```
  DeadUrl: version 14.3.0-r4: HOMEPAGE: No connection adapters were found for 'ttps://characteristic.readthedocs.org/'
  BadHomepage: version 14.3.0-r4: HOMEPAGE='ttps://characteristic.readthedocs.org/' uses unsupported protocol 'ttps'
```

I think we should avoid trying URLs that are unsupported per BadHomepage.

mgorny:
Limit network checks to ftp, http and https URLs, in order to avoid
DeadUrl reports that are duplicate to e.g. BadHomepage.

Fixes #372

### DIFF
--- a/src/pkgcheck/checks/network.py
+++ b/src/pkgcheck/checks/network.py
@@ -208,7 +208,7 @@ class _UrlCheck(NetworkCheck):
             if url.startswith('ftp://'):
                 self._schedule_check(
                     self._ftp_check, attr, url, executor, futures, pkg=pkg)
-            else:
+            elif url.startswith(('https://', 'http://')):
                 self._schedule_check(
                     self._http_check, attr, url, executor, futures, pkg=pkg)
                 http_urls.append((attr, url))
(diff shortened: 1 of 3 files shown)


arthurzam (approved):
@mgorny Looks good, thank you for implementing it.

I do want to request you change the `Fixes` line in commit message to:
`Fixes: https://github.com/pkgcore/pkgcheck/issues/372`

(The full URL is used so when we migrate to gitlab / gitweb, the links will still work as intended)

src/pkgcheck/checks/network.py
```diff
@@ -208,7 +208,7 @@ def schedule(self, pkg, executor, futures):
             if url.startswith('ftp://'):
                 self._schedule_check(
                     self._ftp_check, attr, url, executor, futures, pkg=pkg)
-            else:
+            elif url.startswith('https://') or url.startswith('http://'):
```
radhermit, line 211: str.startswith() can take a tuple of prefixes to check for so I'd simplify it to one call.

arthurzam (approved):

mgorny:
Both done.

PR #387, opened 2022-05-24
add a check for remote-id validity

mgorny:
Add an auxiliary check for invalid remote-id values.  At the moment,
    it primarily checks for the correct number of path components and some
    obvious mistakes.  In the future, the regular expressions can be
    improved to catch invalid user/project names as well.
    
    This includes tests for common valid and invalid values for the current
    set of remote-ids.  In addition to that, PkgMetadataXmlEmptyElement
    check is extended to cover "empty <remote-id/>" path in the code.
    Unfortunately, covering "unknown type" path is non-trivial as it
    requires a remote-id type that is allowed by the XML Schema but not
    covered by validation (which normally happens only when Schema is
    updated).
    
    This also requires bumping the minimal pkgcore version in order to grab
    XML Schema updates (adding "osdn" remote type).

<details><summary>Current output</summary>

```
$ pkgcheck scan -k InvalidRemoteID
app-backup/deja-dup
  InvalidRemoteID: remote-id value 'https://gitlab.gnome.org/World/deja-dup' invalid for type='gitlab', expected: '{username}/[{group}/...]{repo}'

app-crypt/adcli
  InvalidRemoteID: remote-id value 'https://gitlab.freedesktop.org/realmd/adcli' invalid for type='gitlab', expected: '{username}/[{group}/...]{repo}'

app-emulation/wine-gecko
  InvalidRemoteID: remote-id value 'wine/wine-gecko' invalid for type='sourceforge', expected: '{project}'

app-text/pdfgrep
  InvalidRemoteID: remote-id value 'pdfgrep' invalid for type='gitlab', expected: '{username}/[{group}/...]{repo}'

dev-go/gopls
  InvalidRemoteID: remote-id value 'golang/tools/gopls' invalid for type='github', expected: '{username}/{project}'

dev-java/fec
  InvalidRemoteID: remote-id value 'onionnetworks' invalid for type='bitbucket', expected: '{username}/{project}'

dev-java/jackson-dataformat-yaml
  InvalidRemoteID: remote-id value 'https://github.com/FasterXML/jackson-dataformats-text/issues' invalid for type='github', expected: '{username}/{project}'

dev-java/junitparams
  InvalidRemoteID: remote-id value 'Pragmatists/JUnitParams/tags' invalid for type='github', expected: '{username}/{project}'

dev-libs/thrift
  InvalidRemoteID: remote-id value 'apache/thrift/' invalid for type='github', expected: '{username}/{project}'

dev-php/webmozart-assert
  InvalidRemoteID: remote-id value 'webmozart/' invalid for type='github', expected: '{username}/{project}'

dev-python/clikit
  InvalidRemoteID: remote-id value 'clikit' invalid for type='github', expected: '{username}/{project}'

dev-python/statsmodels
  InvalidRemoteID: remote-id value 'statsmodels' invalid for type='github', expected: '{username}/{project}'

dev-python/translate-toolkit
  InvalidRemoteID: remote-id value 'translate' invalid for type='github', expected: '{username}/{project}'

dev-scheme/racket
  InvalidRemoteID: remote-id value 'racket' invalid for type='github', expected: '{username}/{project}'

media-radio/js8call
  InvalidRemoteID: remote-id value 'js8call' invalid for type='bitbucket', expected: '{username}/{project}'

net-analyzer/gvm
  InvalidRemoteID: remote-id value 'greenbone' invalid for type='github', expected: '{username}/{project}'

sci-libs/symmetrica
  InvalidRemoteID: remote-id value 'https://gitlab.com/sagemath/symmetrica' invalid for type='gitlab', expected: '{username}/[{group}/...]{repo}'

x11-themes/ubuntu-wallpapers
  InvalidRemoteID: remote-id value 'ubuntu/+source/ubuntu-wallpapers' invalid for type='launchpad', expected: '{project}'
```
</details>


mgorny:
* figured out how to make gentoo work with a regexp (without extra class)
* added check for `lp:` prefix in launchpad names

testdata/repos/standalone/PackageMetadataXmlCheck/InvalidRemoteID/metadata.xml
```diff
@@ -0,0 +1,15 @@
...
+	<maintainer type="person">
+		<email>maintainer@random.email</email>
+	</maintainer>
+	<upstream>
+		<remote-id type="bitbucket">foo</remote-id>
+		<remote-id type="bitbucket">/foo</remote-id>
+		<remote-id type="bitbucket">foo/</remote-id>
+		<remote-id type="bitbucket">foo/bar/</remote-id>
+		<remote-id type="bitbucket">foo/bar/baz</remote-id>
+		<remote-id type="bitbucket">https://bitbucket.org/foo/bar</remote-id>
```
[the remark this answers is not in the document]
arthurzam: > does this look like a sufficient test set for remotes of `{username}/{project}` form?

Yes, I think this is enough for bitbucket.

> Also, I'm wondering if I should include "good" entries here too, or make them as a "fix" instead. I suppose including them here would be simpler and smaller.

For the bitbucket entries, I don't see any useful testing if we use a `fix.patch`.

testdata/repos/standalone/PackageMetadataXmlCheck/InvalidRemoteID/metadata.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="person">
+		<email>maintainer@random.email</email>
+	</maintainer>
+	<upstream>
```
arthurzam, line 7: All those remotes test very nicely the `_two_components_validator_re` regex. But we are missing for all other regexes. Let's add some failing examples (in this file) for all other regexes.

src/pkgcheck/checks/metadata_xml.py
```diff
@@ -467,6 +514,15 @@ def _check_longdescription(self, pkg, loc, doc):
                 msg = 'metadata.xml longdescription is too short'
                 yield RedundantLongDescription(msg, pkg=pkg)
 
+    def _check_remote_id(self, pkg, loc, doc):
+        for u in pkg.upstreams:
+            try:
+                validator, expected = self.remote_id_validators[u.type]
+            except KeyError:
+                continue
```
arthurzam, line 522: Do I understand correctly that this applies for when the remote-id type is incorrect?
If yes, should we throw another Error/Warning? Also maybe we should add it into `testdata/repos/standalone/PackageMetadataXmlCheck/InvalidRemoteID`?

testdata/repos/standalone/PackageMetadataXmlCheck/InvalidRemoteID/metadata.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="person">
+		<email>maintainer@random.email</email>
```
arthurzam, line 5: Nit pick: we are trying to increase his presence in pkgcheck, so let's use here `larry@example.org`

arthurzam (changes requested):
I think we should also add a `testdata/repos/standalone/PackageMetadataXmlCheck/ValidRemoteID/metadata.xml` which will hold valid remotes (one for each remote).

testdata/repos/standalone/PackageMetadataXmlCheck/InvalidRemoteID/metadata.xml
mgorny: I said this was WIP. I've finished adding all examples, now grouped into 1/2/2+ component cases.

src/pkgcheck/checks/metadata_xml.py
mgorny: Remote-id types are checked through XML Schema. This is primarily meant to allow adding new types without having to update the check immediately.

testdata/repos/standalone/PackageMetadataXmlCheck/InvalidRemoteID/metadata.xml
mgorny: Done.

mgorny:
Ok, I don't think we can cover the `KeyError` case as that would require having an XML Schema entry without a validator.

mgorny:
Ah, right, need to bump pkgcore dep to get fresh XML Schema.

mgorny:
Ok, I think it's ready now.

arthurzam (approved):
Looks great. Thank you!

Issue #385, opened 2022-05-24
metadata.xml --net check results are sometimes not displayed


mgorny:
For example, if I modify app-admin/rex per the following patch:

<details><summary>patch</summary>
```diff
diff --git a/app-admin/rex/metadata.xml b/app-admin/rex/metadata.xml
index ec96a5449c26..81cc5d887b01 100644
--- a/app-admin/rex/metadata.xml
+++ b/app-admin/rex/metadata.xml
@@ -11,348 +11,6 @@
 	</maintainer>
 	<upstream>
 		<remote-id type="github">RexOps/Rex</remote-id>
-		<remote-id type="cpan">Rex</remote-id>
-		<remote-id type="cpan-module">Rex</remote-id>
-		<remote-id type="cpan-module">Rex::Args</remote-id>
-		<remote-id type="cpan-module">Rex::Args::Integer</remote-id>
-		<remote-id type="cpan-module">Rex::Args::Single</remote-id>
-		<remote-id type="cpan-module">Rex::Args::String</remote-id>
-		<remote-id type="cpan-module">Rex::Batch</remote-id>
-		<remote-id type="cpan-module">Rex::Box</remote-id>
-		<remote-id type="cpan-module">Rex::Box::Amazon</remote-id>
-		<remote-id type="cpan-module">Rex::Box::Base</remote-id>
-		<remote-id type="cpan-module">Rex::Box::Docker</remote-id>
-		<remote-id type="cpan-module">Rex::Box::KVM</remote-id>
-		<remote-id type="cpan-module">Rex::Box::VBox</remote-id>
-		<remote-id type="cpan-module">Rex::CLI</remote-id>
-		<remote-id type="cpan-module">Rex::CMDB</remote-id>
-		<remote-id type="cpan-module">Rex::CMDB::Base</remote-id>
-		<remote-id type="cpan-module">Rex::CMDB::YAML</remote-id>
-		<remote-id type="cpan-module">Rex::Cloud</remote-id>
-		<remote-id type="cpan-module">Rex::Cloud::Amazon</remote-id>
-		<remote-id type="cpan-module">Rex::Cloud::Base</remote-id>
-		<remote-id type="cpan-module">Rex::Cloud::Jiffybox</remote-id>
-		<remote-id type="cpan-module">Rex::Cloud::OpenStack</remote-id>
-		<remote-id type="cpan-module">Rex::Commands</remote-id>
[...]

PR #390, opened 2022-05-25
network: Fix metadata.xml results when old version is present

mgorny:
Apply metadata.xml URL check results to the newest package version
rather than the oldest, in order to fix displaying them when multiple
package versions are available.  Since the result objects are subclasses
of FilteredVersionResult, they are filtered out when they do not apply
to the newest version.

Closes: https://github.com/pkgcore/pkgcheck/issues/385


arthurzam (approved):
Look like it is included inside https://github.com/pkgcore/pkgcheck/pull/392, but I can just merge this one :)

PR #392, opened 2022-05-25
network: Limit checks to newest versions in slot (correctly)

mgorny:
Replace the resulting filter in network checks with source filtering. This has the advantage that the checks are actually done only for the interesting versions, rather than being done (and therefore the requests sent) for old versions, and then discarded. This uses a new source class that is roughly based on newest version filtering but actually works.

### DIFF
--- a/src/pkgcheck/objects.py
+++ b/src/pkgcheck/objects.py
@@ -170,9 +170,7 @@ class _KeywordsLazyDict(_LazyDict):
     @klass.jit_attr
     def filter(self):
         """Mapping of default result filters."""
-        from . import results
-        return ImmutableDict({
-            v: 'latest' for v in self.select(results.FilteredVersionResult).values()})
+        return ImmutableDict()
 
 
 class _ChecksLazyDict(_LazyDict):
--- a/src/pkgcheck/sources.py
+++ b/src/pkgcheck/sources.py
@@ -4,6 +4,7 @@ import os
 from collections import defaultdict, deque
 from collections.abc import Set
 from dataclasses import dataclass
+from itertools import groupby
 from operator import attrgetter
 
 from pkgcore.ebuild.profiles import ProfileError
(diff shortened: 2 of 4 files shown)


src/pkgcheck/objects.py
```diff
@@ -171,8 +171,7 @@ def info(self):
     def filter(self):
         """Mapping of default result filters."""
         from . import results
```
arthurzam, line 173: This look redundent now?

src/pkgcheck/objects.py
```diff
@@ -171,8 +171,7 @@ def info(self):
...
-        return ImmutableDict({
```
arthurzam, line 174: What are the effects of this removal of `latest` filter? Does it affect regular usage?
(Note: I haven't understood till now how `latest` filter works in pkgcheck, so asking just in case)

src/pkgcheck/sources.py
```diff
@@ -59,6 +60,19 @@ def itermatch(self, restrict, sorter=sorted, **kwargs):
...
 
+class LatestVersionRepoSource(RepoSource):
+    """Repo source that returns only the latest non-VCS and VCS slots"""
+
+    def itermatch(self, *args, **kwargs):
+        for _, pkgs in groupby(super().itermatch(*args, **kwargs),
+                               key=lambda pkg: pkg.slotted_atom):
+            best_by_live = {}
+            for pkg in pkgs:
+                best_by_live[pkg.live] = pkg
```
arthurzam, line 71: I think list comprehension will be nice here?
`best_by_live = {pkg.live: pkg for pkg in pkgs}`

src/pkgcheck/sources.py
```diff
@@ -59,6 +60,19 @@ def itermatch(self, restrict, sorter=sorted, **kwargs):
...
+            for pkg in sorted(best_by_live.values()):
+                yield pkg
```
arthurzam, line 73: For better context in case an error occurs:
`yield from sorted(best_by_live.values())`

arthurzam (changes requested):
@mgorny Looks good, some cosmetic changes and a sanity question about `latest` filter

src/pkgcheck/sources.py
mgorny: I didn't think it would work here but indeed it seems to.

mgorny:
Done. Note that I haven't run tests locally, so let's wait for CI.

src/pkgcheck/objects.py
arthurzam: OK, thanks for clarification

arthurzam (approved):
Looks great - thank you @mgorny

Issue #417, opened 2022-07-25
Emacs integration


xgqt:
Hello!

I opened this issue to discuss possibilities of pkgcheck integration into GNU Emacs editing workflow.

For few past months I have been trying to write a plugin for flycheck, a on-the-fly checker (https://github.com/flycheck/flycheck) and finally found a way to integrate the components.

# Flycheck-Pkgcheck

## Code

The current state of flycheck-pkgcheck is as follows:
```lisp
(require 'flycheck)


(flycheck-define-checker pkgcheck
  "A checker for ebuild files."
  :modes 'ebuild-mode
  :predicate (lambda () (buffer-file-name))
  :command ("pkgcheck"
            "scan"
            "--exit="
            "--format={package}-{version}.ebuild:0:{level}:{name}: {desc}"
            "--reporter=FormatReporter"
            "--verbose"
            (eval (buffer-file-name)))
  :error-patterns
  ((info
    line-start (file-name) ":" line ":" "info" ":" (message) line-end)
   (info
    line-start (file-name) ":" line ":" "style" ":" (message) line-end)
   (warning
    line-start (file-name) ":" line ":" "warning" ":" (message) line-end)
   (error
    line-start (file-name) ":" line ":" "error" ":" (message) line-end)))


;;;###autoload
(defun flycheck-pkgcheck-setup ()
  "Flycheck pkgcheck setup."
  (interactive)
  (add-to-list 'flycheck-checkers 'pkgcheck t))

;;;###autoload
(add-hook 'ebuild-mode-hook 'flycheck-pkgcheck-setup)


(provide 'flycheck-pkgcheck)
```
It can be also found under this link: https://dev.gentoo.org/~xgqt/scripts/flycheck-pkgcheck.el

## Showcase

![2022 07 25_08:25:16](https://user-images.githubusercontent.com/47485207/180711859-1700ffe8-8e46-4303-8759-6523e03b91d3.png)

# Questions

## lineno alternative

There are a few hacks in `flycheck-pkgcheck` code, most importantly flycheck requires that the offending line is reported with `line`, so I had to hardcode line 0 (`....ebuild:0:{level}...`).

So, a thing to consider in pkgcheck format API: There is `lineo` but it is not guaranteed that it will return a line number, maybe an new keyword could be added to return `lineno` or `0` (eg. `lineno0`)?

## Hosting

Would pkgcheck take this code if I were to open a pull request?

arthurzam:
Thanks for the work.

[...]

PR #420, opened 2022-07-31
contrib/emacs: add flycheck-pkgcheck.el - Flycheck checker for ebuilds

xgqt:
Bug: https://github.com/pkgcore/pkgcheck/issues/417

Signed-off-by: Maciej Barć <xgqt@gentoo.org>

### DIFF
--- a/contrib/emacs/flycheck-pkgcheck.el
+++ b/contrib/emacs/flycheck-pkgcheck.el
@@ -1,0 +1,81 @@
+;;; flycheck-pkgcheck.el --- Flycheck checker for ebuilds -*- lexical-binding: t -*-
+
+
+;; Copyright (c) 2006-2021, pkgcheck contributors
+
+;; Authors: Maciej Barć <xgqt@gentoo.org>
+;; Keywords: convenience processes tools
+;; Homepage: https://pkgcore.github.io/pkgcheck/
+;; SPDX-License-Identifier: BSD-3-Clause
+;; Package-Requires: ((emacs "24.1") (flycheck "32"))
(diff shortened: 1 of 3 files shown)


thesamesam:
Please cherry pick Arthur's commit as he suggested as well.

xgqt:
Right, it was supposed to be the commit to master, my bad.

xgqt:
Switched to `master`. Arthurzam's commit is there too.

arthurzam (approved):
Thank you, I'm merging it 😄 
Now, please do the changes to `pkgcheck-9999` and open a PR on `::gentoo` (note, this is so I can just do small review this looks nice and good, cause the changes you mentions with compiling lisp are unknown for me, so I can't do it myself).

Thank you for improving for all emacs users (even though I'm from vim camp 😉  )

xgqt:
Oh! I realized just now that there is a problem - app-emacs/flycheck is keyworded only on ~amd64.
This will not affect the live version but for keyworded/stabilized releases this is bad.
But Flycheck should actaually be "allarches" but the GNU Emacs team haven't field any bugs to keyword it yet (which I blame myself for :/).

arthurzam:
@xgqt So, some solutions for now:

1. As you said, to keyword on arches - good idea, but maybe not all arches want this dep tree (like hppa or sparc or ia64 or mips)
2. Mask everywhere the use flag, and unmask on amd64 and other arches that are ready.
3. About the stable/unstable, we can mask the flag for stable profiles, but preferably unmask it soon (so stable pkgcheck users can use emacs integration).

Issue #412, opened 2022-06-28
Safe directory feature in >= git 2.35.2 breaks git integration


thesamesam:
With >= git 2.35.2, the "safe directory" feature prevents use of `pkgdev commit` (and other `pkgcheck` git integration). It's not possible to add it as a safe directory like the error suggests because of https://github.com/pkgcore/pkgcheck/commit/ad207610a254c7a502d5760e2e33cf442448b3c4 (added for #336).

```
~/git/overlay/dev-vcs/git $ pkgdev commit
pkgdev commit: error: failed running git log: fatal: unsafe repository ('/var/db/repos/gentoo' is owned by someone else)
To add an exception for this directory, call:

        git config --global --add safe.directory /var/db/repos/gentoo
```

```
~/git/overlay/dev-vcs/git $ pkgdev commit -v --debug
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/pkgcheck/addons/__init__.py", line 288, in init_addon
    addon = addons_map[cls]
KeyError: <class 'pkgcheck.addons.git.GitAddon'>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/pkgcheck/addons/git.py", line 559, in update_cache
    self.pkg_history(
  File "/usr/lib/python3.10/site-packages/pkgcheck/addons/git.py", line 510, in pkg_history
    for pkg in GitRepoPkgs(repo.location, commit_range, local=local):
  File "/usr/lib/python3.10/site-packages/pkgcheck/addons/git.py", line 207, in __init__
    super().__init__(*args)
  File "/usr/lib/python3.10/site-packages/pkgcheck/addons/git.py", line 133, in __init__
    next(self.git_log)
  File "/usr/lib/python3.10/site-packages/pkgcheck/addons/git.py", line 102, in __next__
    raise GitError(f'failed running git log: {error}')
pkgcheck.addons.git.GitError: failed running git log: fatal: unsafe repository ('/var/db/repos/gentoo' is owned by someone else)
To add an exception for this directory, call:

        git config --global --add safe.directory /var/db/repos/gentoo

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python-exec/python3.10/pkgdev", line 4, in <module>
    scripts.run(basename(__file__))
  File "/usr/lib/python3.10/site-packages/pkgdev/scripts/__init__.py", line 38, in run
[...]

PR #421, opened 2022-08-06
addons/git: fix interaction with git's safe.directory

arthurzam:
When using GitLog class, it runs the git commands with clean git config, (aka it uses an empty one) which results in inability to use pkgcheck scan on overlays (as it tries to access system's gentoo repo, which is owned by root). If the user set safe.directory in his config, it will still not work, as we override user's config.

Fix it by creating a temporary file, which we fill with minimal config which disables safe.directory feature, and passing it as the config file to the git log command.

Resolves: https://github.com/pkgcore/pkgcheck/issues/412

### DIFF
--- a/src/pkgcheck/addons/git.py
+++ b/src/pkgcheck/addons/git.py
@@ -10,6 +10,7 @@ from dataclasses import dataclass
 from datetime import datetime
 from functools import partial
 from itertools import takewhile
+import tempfile
 
 from pathspec import PathSpec
 from pkgcore.ebuild import cpv
@@ -31,10 +32,6 @@ from ..base import PkgcheckUserException
 from ..checks import GitCommitsCheck
 from ..log import logger
 from . import caches
-
-
-# ignore global user and system git config
-NO_CONFIG_ENV = ImmutableDict({'GIT_CONFIG_GLOBAL': '', 'GIT_CONFIG_SYSTEM': ''})
 
 
 @dataclass(frozen=True, eq=False)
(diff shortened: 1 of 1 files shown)


src/pkgcheck/addons/git.py
```diff
@@ -79,14 +76,37 @@ def __init__(self, *args, commit):
         self.commit = commit
 
 
+class GitSafeDirectory:
```
laumann, line 79: [suggestion] This is meant to be used to override whatever configuration the user has, maybe a better name is `GitGlobalConfig` or something to that effect? Or just `GitConfig`.

arthurzam:
@thesamesam does this solution make sense for you?

src/pkgcheck/addons/git.py
```diff
@@ -79,14 +76,37 @@ def __init__(self, *args, commit):
         self.commit = commit
 
 
+class GitConfig:
+    """Manages temporary file which holds git config for disabling
+    safe directory feature of git."""
+
+    def __init__(self):
+        fd, self.path = tempfile.mkstemp()
+        print('AAA', self.path)
```
thesamesam, line 85: ?

laumann: Looks like debugging :-)

src/pkgcheck/addons/git.py
```diff
@@ -79,14 +76,37 @@ def __init__(self, *args, commit):
...
+    def close(self):
+        os.unlink(self.path)
+
+
 class GitLog:
     """Iterator for decoded `git log` line output."""
 
     def __init__(self, cmd, path):
         self._running = False
+        self.safe_directory = GitConfig()
```
laumann, line 106: [nit] I would maybe rename this variables to `self.git_config` so it'd be
```python
env=self.git_config.config_env
```
in the call to `subprocens.Popen`

laumann (commented):
Looks good, the debugging print line should probably go. Nice work.

src/pkgcheck/addons/git.py
arthurzam: Oh, I missed this line.
Yes, this was for debugging the auto delete part.

src/pkgcheck/addons/git.py
arthurzam: done

thesamesam (approved):
We might need to revisit whether we really want to ignore 'git log' in future (probably by revisiting git bindings in future).

Unfortunately, not that long ago at all, when radhermit tried to use the various git Python bindings out there, IIRC they were all far slower and had other limitations. But w/e.

This is what I was expecting and looks good, thanks!

PR #423, opened 2022-08-07
UncheckableDep: add testcase

thesamesam:
UncheckableDep: add testcase

Adapted from dev-lang/rust. We didn't have coverage
over this result at all, so throw it in.

More coverage is always good, but in particular,
we want to know if we stop immediately
aborting on complicated deps. It could be a regression
(we parse for ages and give up -> hang).


arthurzam (approved):
Thanks

PR #429, opened 2022-08-13
StableRequestCheck: Add stabletime config option

EmRowlands:
This commit adds an option allows the user to specify the time before a version is flagged by StableRequestCheck. The primary use case for this is in overlays where the stabilisation policy may be different from ::gentoo.

I deliberately did not add a short option for `--stabletime` because there are a limited number of valid single-letter options, and I didn't think that this case warranted using one. If maintainers think it's a good idea to have one, I'd be happy to add it in.

### DIFF
--- a/src/pkgcheck/scripts/pkgcheck_scan.py
+++ b/src/pkgcheck/scripts/pkgcheck_scan.py
@@ -190,6 +190,14 @@ check_options.add_argument(
         Use ``pkgcheck show --keywords`` to see available options.
     """)
 
+check_options.add_argument(
+    '--stabletime', metavar='DAYS', dest='stable_time', default=30,
+    type=arghparse.positive_int, help='set number of days before stabilisation',
+    docs="""
+        An integer number of days before a package version is flagged by
+        StableRequestCheck. Defaults to 30 days.
+    """)
+
 scan.plugin = scan.add_argument_group('plugin options')
 
 
--- a/tests/checks/test_stablereq.py
+++ b/tests/checks/test_stablereq.py
@@ -34,19 +34,21 @@ class TestStableRequestCheck(ReportTestCase):
         self.child_git_repo.run(['git', 'remote', 'set-head', 'origin', 'main'])
         self.child_repo = make_repo(self.child_git_repo.path)
 
-    def init_check(self, options=None, future=0):
-        self.options = options if options is not None else self._options()
+    def init_check(self, options=None, future=0, stable_time=None):
+        self.options = options if options is not None else self._options(stable_time=stable_time)
         self.check, required_addons, self.source = init_check(self.check_kls, self.options)
         for k, v in required_addons.items():
             setattr(self, k, v)
         if future:
             self.check.today = datetime.today() + timedelta(days=+future)
 
-    def _options(self, **kwargs):
+    def _options(self, stable_time=None, **kwargs):
         args = [
             'scan', '-q', '--cache-dir', self.cache_dir,
             '--repo', self.child_repo.location,
         ]
+        if stable_time is not None:
+            args.extend(['--stabletime', str(stable_time)])
         options, _ = self._tool.parse_args(args)
         return options
 
@@ -73,7 +75,14 @@ class TestStableRequestCheck(ReportTestCase):
         self.init_check(future=30)
         self.assertNoReport(self.check, self.source)
 
-    def test_existing_stable_keywords(self):
+    @pytest.mark.parametrize(("stable_time", "less_days", "more_days"), (
+        pytest.param(None,  (0, 1, 10, 20, 29), (30, 31),     id="stable_time=unset"),
+        pytest.param(1,     (0,),               (1, 10),      id="stable_time=1"),
+        pytest.param(14,    (0, 1, 10, 13),     (14, 15, 30), id="stable_time=14"),
+        pytest.param(30,    (0, 1, 10, 20, 29), (30, 31),     id="stable_time=30"),
+        pytest.param(100,   (98, 99),           (100, 101),   id="stable_time=100"),
+    ))
+    def test_existing_stable_keywords(self, stable_time, less_days, more_days):
         self.parent_repo.create_ebuild('cat/pkg-1', keywords=['amd64'])
         self.parent_git_repo.add_all('cat/pkg-1')
         self.parent_repo.create_ebuild('cat/pkg-2', keywords=['~amd64'])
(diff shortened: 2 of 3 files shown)


thesamesam:
Seems reasonable.

EmRowlands:
Instead of adding a new test, I've added a fixture to an existing test that runs with varying values of `--stabletime`. The existing test case is preserved by including a case that uses a default value.

This solution might be a bit over-engineering-y, so I'm happy to split the default-value case back out into a separate test and use `pytest.mark.parametrize` instead. I decided to use this way round because it allows for other tests to use it in future.

tests/checks/test_stablereq.py
```diff
@@ -34,6 +46,21 @@ def _setup(self, tmp_path, tool, make_repo, make_git_repo):
...
+        ids=map(lambda paramset: f"stable_time={paramset['days'] or 'unset'}", _stable_time_params)
+    )
+    def stable_time(self, request):
+        self.stable_time = request.param["days"]
+        self.less_days = request.param["less"]
+        self.more_days = request.param["more"]
+
+        return request.param
+
     def init_check(self, options=None, future=0):
```
arthurzam, line 64: Let's pass the `stable_time` as an argument here, so less using of `self.`:

```suggestion
    def init_check(self, options=None, future=0, stable_time=None):
```

And then change the `if self.stable_time is not None:` to use the argument directly

tests/checks/test_stablereq.py
```diff
@@ -91,6 +120,26 @@ def test_existing_stable_keywords(self):
         expected = StableRequest('0', ['~amd64'], 30, pkg=VersionedCPV('cat/pkg-2'))
         assert r == expected
 
+    def test_existing_stable_keywords_stable_time(self, stable_time):
```
arthurzam, line 123: I would prefer to use some better format for the parameterized fixture:
```suggestion
    @pytest.mark.parametrize(("stable_days", "less_days", "more_days"), (
        pytest.param(None, (0, 1, 10, 20, 29), (30, 31), id="stable_time=unset"),
        pytest.param(1, (0,), (1, 10), id="stable_time=1"),
        pytest.param(14, (0, 1, 10, 13), (14, 15, 30), id="stable_time=14"),
        pytest.param(30, (0, 1, 10, 20, 29), (30, 31), id="stable_time=30"),
        pytest.param(100, (98, 99), (100, 101), id="stable_time=100"),
    ))
    def test_existing_stable_keywords_stable_time(self, stable_days, less_days, more_days):
```

That way you can remove all the previous fixture, and use direct arguments without polluting `self.` I also think it looks more clean as a fixture list.

arthurzam (commented):
@EmRowlands Could you please add a signoff to the commit (to the same rules as for Gentoo repository)?
While this isn't a must policy, I do plan to move the whole stack to Gentoo infra repo, and having a signoff would make our life easier.

---

Also, this looks very nice, so I do plan to merge it after those comments. In which case during the release commit, I need to write your contribution in the NEWS file with your name (as a thanks token) - so which is the preferred way to write your full name?

src/pkgcheck/scripts/pkgcheck_scan.py
```diff
@@ -190,6 +190,14 @@
         Use ``pkgcheck show --keywords`` to see available options.
     """)
 
+check_options.add_argument(
+    '--stabletime', metavar='DAYS', dest='stable_time', default=30,
+    type=arghparse.positive_int, help='set number of days before stabilisation',
+    docs="""
+        An integer number of days before a package version is flagged by
+        StableRequestCheck. Defaults to 30 days.
```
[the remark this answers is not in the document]
thesamesam: We could say "whole number" instead of "integer" if you prefer, but it reads okay as it is.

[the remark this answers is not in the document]
arthurzam: OK, then ignore this part - good as it is now 😄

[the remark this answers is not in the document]
EmRowlands: Now that you point it out, I agree with you: "integer" is redundant here.

tests/checks/test_stablereq.py
EmRowlands: This makes much more sense, I'll implement this.

tests/checks/test_stablereq.py
EmRowlands: As I mentioned in https://github.com/pkgcore/pkgcheck/pull/429#issuecomment-1214202815, I thought I might have gone overboard. I'll parameterize the test

EmRowlands:
> Could you please add a signoff to the commit (to the same rules as for Gentoo repository)? While this isn't a must policy, I do plan to move the whole stack to Gentoo infra repo, and having a signoff would make our life easier.

Last I checked, I was unable to contribute to Gentoo itself because I ran afoul of one of the relevant GLEPs. I'll have a read through of the notes I made after I push the changes you've suggested, and I'll push again later with the signoff if there's no issues.

> Also, this looks very nice, so I do plan to merge it after those comments. In which case, I need to write your contribution in the NEWS file with your name (as a thanks token) - so which is the preferred way to write your full name?

Thank you, I appreciate the gesture. My full name can be written as it appears in the git log: "Emily Rowlands".

tests/checks/test_stablereq.py
EmRowlands: Implemented in [16aeb72](https://github.com/pkgcore/pkgcheck/pull/429/commits/16aeb7281e82bb7ae0801f7589d0ce740c7ca44f)

tests/checks/test_stablereq.py
EmRowlands: Implemented in [16aeb72](https://github.com/pkgcore/pkgcheck/pull/429/commits/16aeb7281e82bb7ae0801f7589d0ce740c7ca44f)

EmRowlands:
I've implemented the suggested changes, but (as I said in https://github.com/pkgcore/pkgcheck/pull/429#issuecomment-1214217906) I've not added the signoff.

I'll report back once I've investigated why it was that I was previously unable to do it.

thesamesam:
Please feel free to email me privately (or arthur) if needed.

EmRowlands:
(force-push again because I messed up my rebase)

EmRowlands:
Good news: the situation that prevented me from doing the signoff no longer applies.

Please re-review or merge if ready!

arthurzam (approved):

arthurzam:
> > Also, this looks very nice, so I do plan to merge it after those comments. In which case, I need to write your contribution in the NEWS file with your name (as a thanks token) - so which is the preferred way to write your full name?
> 
> Thank you, I appreciate the gesture. My full name can be written as it appears in the git log: "Emily Rowlands".

@EmRowlands, as agreed, you can see your name in NEWS file, and in [v0.10.13 release notes](https://github.com/pkgcore/pkgcheck/releases/tag/v0.10.13)
Thank you very much for the contribution, and we would be thankful for any future contributions!

Issue #399, opened 2022-05-30
Detect pure-DEPEND weak blockers (!)


thesamesam:
Example: https://github.com/gentoo/gentoo/commit/6b2955f24cb027e3aa10e6dfde91d792a4b08a5c.

Pure-DEPEND weak blockers (!) don't work as they control whether we ignore file collisions at the point of merge - hence they belong in RDEPEND. It's fine but not necessary for them to be in both DEPEND and RDEPEND.

Such cases should be changed to RDEPEND or RDEPEND +DEPEND (same effect as just RDEPEND)

arthurzam:
OK, so just to check I understand correctly:

Usage of Pure-DEPEND weak blockers (those that use single `!`) in those types:

1. RDEPEND - fine and the correct place
2. (BDEPEND or DEPEND) but not in RDEPEND - wrong (most likely mistake) and should show a warning, or even error? (after gentoo tree is fixed)
3. (BDEPEND or DEPEND) but also in RDEPEND - wrong, should show warning that it is useless in first group
4. IDEPEND - I think fine, but really not sure...

PR #430, opened 2022-08-14
new check: MisplacedWeakBlocker

arthurzam:
Resolves: https://github.com/pkgcore/pkgcheck/issues/399

### DIFF
--- a/src/pkgcheck/checks/metadata.py
+++ b/src/pkgcheck/checks/metadata.py
@@ -804,6 +804,27 @@ class InvalidIdepend(results.MetadataError, results.VersionResult):
     attr = 'idepend'
 
 
+class MisplacedWeakBlocker(results.Warning, results.VersionResult):
+    """Weak blocker is within a misplaced dependency class.
+
+    Weak blockers control whether we ignore file collisions at the point of
+    merge, so being (exclusively) defined in DEPEND or BDEPEND is wrong.
+
+    Note that in cases where the weak blocker is also defined in RDEPEND, this
+    warning won't be triggered, to give leeway given this is a common ebuild
+    pattern.
+    """
+
+    def __init__(self, attr, atom, **kwargs):
+        super().__init__(**kwargs)
+        self.attr = attr.upper()
+        self.atom = str(atom)
+
+    @property
+    def desc(self):
+        return f'{self.attr}: misplaced weak blocker: {self.atom}'
+
+
 class DependencyCheck(Check):
     """Verify dependency attributes (e.g. RDEPEND)."""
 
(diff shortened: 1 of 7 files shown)


src/pkgcheck/checks/metadata.py
```diff
@@ -804,14 +804,34 @@ class InvalidIdepend(results.MetadataError, results.VersionResult):
     attr = 'idepend'
 
 
+class MisplacedWeakBlocker(results.Warning, results.VersionResult):
+    """Weak blocker is in misplaced depends type.
```
thesamesam, line 808: Weak blocker is within a misplaced dependency eclass.

src/pkgcheck/checks/metadata.py
```diff
@@ -804,14 +804,34 @@ class InvalidIdepend(results.MetadataError, results.VersionResult):
...
+
+    Weak blockers control whether we ignore file collisions at the point of
+    merge, so being defined at DEPEND or BDEPEND is wrong.
```
thesamesam, line 811:
```suggestion
    merge, so being (exclusively) defined in DEPEND or BDEPEND is wrong.
```

src/pkgcheck/checks/metadata.py
```diff
@@ -804,14 +804,34 @@ class InvalidIdepend(results.MetadataError, results.VersionResult):
...
+
+    Note that in cases where the weak blocker is also defined at RDEPEND, this
```
thesamesam, line 813:
```suggestion
    Note that in cases where the weak blocker is also defined in RDEPEND, this
```

src/pkgcheck/checks/metadata.py
```diff
@@ -804,14 +804,34 @@ class InvalidIdepend(results.MetadataError, results.VersionResult):
...
+    warning won't be triggered, as a leeway.
```
thesamesam, line 814:
```suggestion
    warning won't be triggered, to give leeway given this is a common ebuild pattern.
```

src/pkgcheck/checks/metadata.py
```diff
@@ -889,6 +911,14 @@ def feed(self, pkg):
                         elif atom.slot_operator == '=':
                             yield BadDependency(
                                 attr, atom, '= slot operator used in blocker', pkg=pkg)
+                        elif not atom.blocks_strongly:
+                            weak_blocks[attr].add(atom)
+
+        for attr in ('depend', 'bdepend'):
```
thesamesam, line 917: Should we add IDEPEND in this block somewhere?

arthurzam: Will handle it tomorrow.
I was thinking that having a weak-blocker in `IDEPEND` was fine, but on second thought, why should it?

thesamesam: It's okay if it's in there along with other deps (but unlikely), but it doesn't make sense anyway, and it's like putting it in RDEPEND. We should treat it like the "In RDEPEND but also DEPEND" case.

arthurzam: Done, thanks for the clarification 😄

thesamesam (approved):

PR #435, opened 2022-08-26
RdependChange: don't report when USE flags are added or removed

arthurzam:


thesamesam:
Note that when _adding_ a USE flag **where the functionality was previously on** (not off in build system), it's valid to report RdependChange, because a dep was unconditional that now became conditional.

Also, if _removing_ a USE flag **where the behaviour is now on, but was previously always off** (not on in build system), it's valid to report RdependChange also, because people might have never set USE=foo (hence --changed-use wouldn't pick it up) but their deps have now effectively changed

arthurzam:
Hmm, I see that I misunderstood a little this Changed Use facility. I'm closing this PR.

Maybe it would be worth to add this explanation to the devmanual somehow, as this was new to me, and I have read that page a lot.
I guess just [that line](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html) is not enough:

> adding a new USE flag or removing an existing one (since change in USE flags is going to trigger --changed-use rebuild),

thesamesam:
Filed https://github.com/pkgcore/pkgcheck/pull/435.

PR #438, opened 2022-09-02
RedundantVersionCheck: add `--stable-only` option

arthurzam:
Add an option for considering only stable versions when checking for redundant versions.
This is useful for cases of cleanup after successful stabilization.

My logic is: for each redundant version, check if it is redundant by a ebuild which has a stable keyword

### DIFF
--- a/src/pkgcheck/checks/cleanup.py
+++ b/src/pkgcheck/checks/cleanup.py
@@ -33,6 +33,17 @@ class RedundantVersionCheck(Check):
     _source = sources.PackageRepoSource
     known_results = frozenset([RedundantVersion])
 
+    @staticmethod
+    def mangle_argparser(parser):
+        parser.plugin.add_argument(
+            '--stable-only', action='store_true',
+            help='consider redundant versions only within stable',
+            docs="""
+                If enabled, for each slot, only consider redundant versions
+                with stable keywords. This is useful for cases of cleanup after
+                successful stabilization.
+            """)
+
     def feed(self, pkgset):
         if len(pkgset) == 1:
             return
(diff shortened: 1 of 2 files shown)


arthurzam:
@thesamesam If you can look at this idea and help texts?

src/pkgcheck/checks/cleanup.py
```diff
@@ -33,6 +33,17 @@ class RedundantVersionCheck(Check):
     _source = sources.PackageRepoSource
     known_results = frozenset([RedundantVersion])
 
+    @staticmethod
+    def mangle_argparser(parser):
+        parser.plugin.add_argument(
+            '--by-stable', action='store_true',
```
thesamesam, line 39: Not sure, but maybe:
```suggestion
            '--stable-only', action='store_true',
```

src/pkgcheck/checks/cleanup.py
```diff
@@ -33,6 +33,17 @@ class RedundantVersionCheck(Check):
...
+            help='consider redundant version by stable versions only',
```
thesamesam, line 40:
```suggestion
            help='consider redundant versions only within stable',
```

src/pkgcheck/checks/cleanup.py
```diff
@@ -33,6 +33,17 @@ class RedundantVersionCheck(Check):
...
+            docs="""
+                If enabled, only consider for each slot redundant versions by
```
thesamesam, line 42:
```suggestion
                If enabled, for each slot, only consider redundant versions with stable keywords.
```

thesamesam (approved):
I like it, thank you!

thesamesam (approved):

PR #446, opened 2022-09-16
network: Add `savannah` and `savannah-nongnu` remote-ids

thesamesam:


arthurzam (approved):
Thanks

Issue #326, opened 2021-05-28
pkgcheck scan --commit: IndexError: pop from an empty deque


mgorny:
```
pkgcheck scan: error: Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/pkgcheck/pipeline.py", line 186, in _run_checks
    if results := sorted(chain.from_iterable(
  File "/usr/lib/python3.9/site-packages/pkgcheck/runners.py", line 67, in run
    for item in self.source.itermatch(restrict):
  File "/usr/lib/python3.9/site-packages/pkgcheck/sources.py", line 36, in itermatch
    yield from self.source
  File "/usr/lib/python3.9/site-packages/pkgcheck/addons/git.py", line 183, in __next__
    pkgs = list(chain.from_iterable(pkgs for _, pkgs in self.changes))
  File "/usr/lib/python3.9/site-packages/pkgcheck/addons/git.py", line 183, in <genexpr>
    pkgs = list(chain.from_iterable(pkgs for _, pkgs in self.changes))
  File "/usr/lib/python3.9/site-packages/pkgcheck/addons/git.py", line 157, in changes
    path = changes.popleft()
IndexError: pop from an empty deque
```

If I debug-print the `changes` deque before the loop, I get:

```
deque(['M', 'sys-kernel/gentoo-kernel-bin/Manifest', 'A', 'sys-kernel/gentoo-kernel-bin/gentoo-kernel-bin-5.12.8.ebuild'])
deque(['M', 'sys-kernel/gentoo-kernel-bin/Manifest', 'A', 'sys-kernel/gentoo-kernel-bin/gentoo-kernel-bin-5.10.41.ebuild'])
deque(['M', 'sys-kernel/gentoo-kernel-bin/Manifest', 'A', 'sys-kernel/gentoo-kernel-bin/gentoo-kernel-bin-5.4.123.ebuild'])
deque(['A', 'virtual/dist-kernel/dist-kernel-5.12.8.ebuild'])
deque(['M', 'sys-kernel/gentoo-kernel/Manifest', 'A', 'sys-kernel/gentoo-kernel/gentoo-kernel-5.12.8.ebuild'])
deque(['M', 'sys-kernel/vanilla-kernel/Manifest', 'A', 'sys-kernel/vanilla-kernel/vanilla-kernel-5.12.8.ebuild'])
deque(['A', 'virtual/dist-kernel/dist-kernel-5.10.41.ebuild'])
[...]

PR #454, opened 2022-09-30
Fixes for `pkgcheck scan --commit`

arthurzam:
Add a special runner which is run only on main process, and after the synchronous runners. This is useful for runners which must not be run in parallel, like the git checks.

Resolves: https://github.com/pkgcore/pkgcheck/issues/326


arthurzam:
I hope this indeed fixes the issue, but for checks, I will call other devs to test it (after applying):

Before pushing your changes, please run `pkgcheck scan --commits` multiple times without any `-j`, preferably on many core systems (>=8 cores).

PR #458, opened 2022-10-05
EbuildReservedCheck: catch declaration of phase hooks

arthurzam:
Phase hooks (`{pre,post}_${phase}`) are used by portage to run user's hooks defined at `bashrc` file. Defining them in ebuilds might cause unexpected behavior, so we should warn about it.

In gentoo tree, only one such ebuild was caught:
```
dev-util/idea-community
  EbuildReservedName: version 2022.2.2: line 41: funciton name "post_src_unpack" is disallowed because "post_src_unpack" is a reserved phase hook
```

Also, in this PR I fix the line number reported, as it was off by one.


thesamesam (approved):
It's arguably not reserved but in reality, let's say it is. Thank you!

PR #464, opened 2022-10-09
contrib/emacs: run flycheck only when buffer is saved

parona-source:
* Instead of checking buffer-file-name use built-in function flycheck-buffer-saved-p which requires the buffer to be saved. Prior it would lead to the check being ran on an non-existent file when initially creating an ebuild file.

* Also remove --verbose as listing out every profile is noise that hinders using the check especially when hovering over with a mouse cursor.

Signed-off-by: Alfred Wingate <parona@protonmail.com>

### DIFF
--- a/contrib/emacs/flycheck-pkgcheck.el
+++ b/contrib/emacs/flycheck-pkgcheck.el
@@ -1,14 +1,15 @@
 ;;; flycheck-pkgcheck.el --- Flycheck checker for ebuilds -*- lexical-binding: t -*-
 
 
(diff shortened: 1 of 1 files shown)


arthurzam:
@xgqt because I don't understand at all what is going on with flycheck, I leave it to you 😄 

This looks fine for me, so your call is enough.

xgqt:
I have tested this and it works correctly. 
Thanks @parona-source !
Sorry for the delay, I forgot about this PR. :sweat:

arthurzam (approved):

Issue #467, opened 2022-10-12
Warn about non-PEP517 distutils-r1 ebuilds


mgorny:
I'd like to prevent people from adding new non-PEP517 ebuilds unless absolutely necessary. I think a good step towards that would be having pkgcheck warn about non-PEP517 mode being deprecated.

I think the simplest way of doing that would be to check for distutils-r1 inherit with no `DISTUTILS_USE_PEP517` declaration.

mgorny:
Ok, I'm going to give it a shot today.

PR #468, opened 2022-10-13
PythonCheck: warn about use of distutils-r1 non-PEP517 mode

mgorny:
Closes: https://github.com/pkgcore/pkgcheck/issues/467

### DIFF
--- a/src/pkgcheck/checks/python.py
+++ b/src/pkgcheck/checks/python.py
@@ -139,6 +139,17 @@ class PythonEclassError(results.VersionResult, results.Error):
         return self.msg
 
 
+class DistutilsNonPEP517Build(results.VersionResult, results.Warning):
+    """Ebuild uses the deprecated non-PEP517 build"""
+
+    @property
+    def desc(self):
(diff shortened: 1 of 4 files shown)


src/pkgcheck/checks/python.py
```diff
@@ -240,6 +252,20 @@ def check_pep517_depend(self, item):
 
         return True
 
+    def check_pep517_mode(self, item):
```
arthurzam, line 255: What do you think about merging the code here with `check_pep517_depend` and rename the function to `check_pep517`?

Then instead of using `return {True,False}`, use `yield` of corresponding result and when calling the function (after if of inherit class) do `yield from self.check_pep517(...)`?

We have very similar code between the two, and is somewhat sad we traverse the tree-sitter query 2 times.

arthurzam (changes requested):
Please ignore the coverage, I think I broke it because on master it has x2 of reports (one from original PR and one from master).

arthurzam:
Merged

PR #484, opened 2022-10-27
VirtualProvidersCheck: new check for providers issues

arthurzam:
- check for virtual package defining DEPEND or BDEPEND
- check for virtual package with a single provider across versions
- All of them of `Warning` level

The logic for the single provider is:

1. Every version has one provider (compares full atom, for cases like when we depend on one of use flags)
2. Collecting all unversioned unslotted RDEPENDs gives us only one match --> to which we should replace the virtual.

Closes: https://bugs.gentoo.org/744784

<details>
  <summary>results over gentoo repo</summary>
  
  ```
  virtual/acl
    VirtualWithSingleProvider: virtual package with a single provider: sys-apps/acl
  
  virtual/blas
    VirtualWithDepend: version 3.8: virtual package with a DEPEND defined
  
  virtual/cblas
    VirtualWithDepend: version 3.8: virtual package with a DEPEND defined
  
  virtual/imap-c-client
    VirtualWithSingleProvider: virtual package with a single provider: net-libs/c-client
  
  virtual/jpeg
    VirtualWithSingleProvider: virtual package with a single provider: media-libs/libjpeg-turbo
  
  virtual/lapacke
    VirtualWithDepend: version 3.8-r1: virtual package with a DEPEND defined
  
  virtual/lapack
    VirtualWithDepend: version 3.8: virtual package with a DEPEND defined
    VirtualWithDepend: version 3.10: virtual package with a DEPEND defined
  
  virtual/libiconv
    VirtualWithSingleProvider: virtual package with a single provider: dev-libs/libiconv
  
  virtual/libintl
    VirtualWithSingleProvider: virtual package with a single provider: dev-libs/libintl
  
  virtual/linuxtv-dvb-headers
    VirtualWithSingleProvider: virtual package with a single provider: sys-kernel/linux-headers
  
  virtual/perl6
    VirtualWithSingleProvider: virtual package with a single provider: dev-lang/rakudo
  
  virtual/pkgconfig
    VirtualWithSingleProvider: virtual package with a single provider: dev-util/pkgconf
  
  virtual/ruby-ssl
    VirtualWithDepend: version 11: virtual package with a DEPEND defined
    VirtualWithBdepend: version 12: virtual package with a BDEPEND defined
    VirtualWithDepend: version 12: virtual package with a DEPEND defined
  
  virtual/rubygems
    VirtualWithBdepend: version 16: virtual package with a BDEPEND defined
    VirtualWithDepend: version 16: virtual package with a DEPEND defined
    VirtualWithBdepend: version 17: virtual package with a BDEPEND defined
    VirtualWithDepend: version 17: virtual package with a DEPEND defined
  
  virtual/tex-base
    VirtualWithSingleProvider: virtual package with a single provider: app-text/texlive-core
  ```
</details>

---

- [ ] Handle the case of `virtual/libiconv` and `virtual/libintl`

### DIFF
--- a/src/pkgcheck/checks/metadata.py
+++ b/src/pkgcheck/checks/metadata.py
@@ -7,7 +7,7 @@ from difflib import SequenceMatcher
 from functools import partial
 from operator import attrgetter
 
-from pkgcore.ebuild import atom as atom_mod
+from pkgcore.ebuild import atom as atom_mod, restricts
 from pkgcore.ebuild.atom import atom as atom_cls
 from pkgcore.ebuild.eapi import get_eapi
 from pkgcore.ebuild.misc import sort_keywords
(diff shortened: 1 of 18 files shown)


thesamesam:
>- check for virtual package defining DEPEND or RDEPEND

You mean BDEPEND or DEPEND? ("not RDEPEND")?

src/pkgcheck/checks/metadata.py
```diff
@@ -1627,3 +1627,67 @@ def feed(self, pkg):
         for unpackers, filenames in missing_unpackers.items():
             yield MissingUnpackerDep(
                 str(pkg.eapi), sorted(filenames), sorted(unpackers), pkg=pkg)
+
+
+class VirtualWithSingleProvider(results.PackageResult, results.Warning):
```
thesamesam, line 1632: _Maybe_ we should suppress this if the package is in `package.deprecated`? 

Not sure if it's worth the effort.

thesamesam:
These are false positives in an interesting way:
```
virtual/libiconv
  VirtualWithSingleProvider: virtual package with a single provider: dev-libs/libiconv

virtual/libintl
  VirtualWithSingleProvider: virtual package with a single provider: dev-libs/libintl
```

In libintl:
```
# - Don't put elibc_glibc? ( sys-libs/glibc ) to avoid circular deps between that and gcc.
RDEPEND="!elibc_glibc? ( !elibc_musl? ( dev-libs/libintl[${MULTILIB_USEDEP}] ) )"
```

The reason is that non-glibc & non-musl usually need an external libintl. In fact, musl needed it for a long time. This lets ebuilds be correct when they need libintl.

A real example of this is macOS, btw. Same for libiconv.

src/pkgcheck/checks/metadata.py
arthurzam: done

arthurzam:
> I'm not sure yet how to cleanly identify these. Maybe if the RDEPEND is "complex" (USE-conditionals) for VirtualWithSingleProvider, we should skip?

Added exception check for dependence on use flags starting with `elibc_` and `kernel_`.

thesamesam (approved):
Sorry, I just wanted to check one thing (for other USE_EXPANDs), and we're all good.

Issue #477, opened 2022-10-18
Warn on non-portable compressed `doman` and `newman` arguments


thesamesam:
See https://bugs.gentoo.org/619916. PMS does not state that compressed `doman` and `newman` arguments must be accepted, hence relying on them in ebuilds is nonportable.

Portage warns on this (this example is at `68c2e4bfc966c7ef2154474a14c7bdc338fd53cb`, [app-misc/socnetv-2.2](https://github.com/gentoo/gentoo/blob/68c2e4bfc966c7ef2154474a14c7bdc338fd53cb/app-misc/socnetv/socnetv-2.2.ebuild)):
```
 * QA Notice: doman argument '/var/tmp/portage/app-misc/socnetv-2.2/work/socnetv-2.2/man/socnetv.1.gz' is compressed, this is not portable
>>> Completed installing app-misc/socnetv-2.2 into /var/tmp/portage/app-misc/socnetv-2.2/image

 * Final size of build directory: 138300 KiB (135.0 MiB)
 * Final size of installed tree:   31696 KiB ( 30.9 MiB)

 * QA Notice: One or more compressed files were found in docompress-ed
 * directories. Please fix the ebuild not to install compressed files
 * (manpages, documentation) when automatic compression is used:
 *
 *   /usr/share/man/man1/socnetv.1.gz
```
but pkgcheck should be able to detect it statically.

Example bad calls:
```
src_install() {
	[...]

	doman foo.1.gz # .gz should not be here!
	newman foo.1.gz <anything> # .gz should not be in source!
}
```

PR #485, opened 2022-10-28
InstallCompressedManpageCheck: new check

arthurzam:
New check for catching passing compressed manpages to doman and newman.

No catches in gentoo repo, but for the one ebuild mentioned in issue, it looks like:
```
app-misc/socnetv
  InstallCompressedManpage: version 2.2: line 41: compressed manpage '${S}/man/${PN}.1.gz' passed to doman
```

Resolves: https://github.com/pkgcore/pkgcheck/issues/477


src/pkgcheck/checks/codingstyle.py
```diff
@@ -1118,3 +1118,33 @@ def feed(self, pkg):
...
+
+    def feed(self, pkg):
+        for node, _ in bash.cmd_query.captures(pkg.tree.root_node):
+            call_name = pkg.node_str(node.child_by_field_name('name'))
+            if call_name not in ('doman', 'newman'):
+                continue
+            for arg in node.children[1:]:
+                arg_name = pkg.node_str(arg).strip('\'\"')
+                lineno, _ = arg.start_point
+                if arg_name.endswith(('.gz', '.bz2', '.xz')):
```
thesamesam, line 1149: Technically the list could be much bigger, but it's quite unlikely upstreams would ship them with any other suffix.

The list at https://github.com/gentoo/portage/commit/636dd15437eccfaf63e76c36412999b5e118c209 is mostly complete fwiw.

arthurzam: nice, thanks!

thesamesam:
I wonder if we should cover `doinfo` too.

arthurzam:
> > I wonder if we should cover `doinfo` too.
> 
> Given how comparatively rare info pages are and I've never seen a compressed one from upstream, let's leave it.

lol, by the time you wrote it, the implementation was already done :)

src/pkgcheck/checks/codingstyle.py
```diff
@@ -1118,3 +1118,53 @@ def feed(self, pkg):
...
+        super().__init__(**kwargs)
+        self.func = func
+
+    @property
+    def desc(self):
+        return f'line {self.lineno}: compressed manpage {self.line!r} passed to {self.func}'
+
+
+class InstallCompressedInfo(results.LineResult, results.Warning):
+    """Compressed manpages are not supported by ``doinfo`` or ``newinfo``."""
```
thesamesam, line 1136: newinfo doesn't exist!

thesamesam (approved):
LGTM once newinfo removed.

src/pkgcheck/checks/codingstyle.py
arthurzam: lol - Just shows how much I don't use it 😄 
(FYI, I even have in my system `INSTALL_MASK="/usr/share/info/"`)

Issue #475, opened 2022-10-18
Detect missing remote-ids based on SRC_URI


thesamesam:
pkgcheck could detect missing remote-ids in metadata.xml by checking SRC_URI in ebuilds.

For example, if an ebuild has `SRC_URI="https://github.com/foo/bar/..."` but no remote-id in metadata.xml, we should warn about a missing element in there: `<remote-id type="github">foo/bar</remote-id>`.

arthurzam:
Some notes for myself:

Reusing the definitions under `MetadataUrlCheck` won't be possible, as various SRC_URI have separate URL scheme than the one for remote-id (for example pypi). This means this check will need separate dict and extraction value for remote-id.

I want to minimize false positive, so some extra rules:
- Don't suggest anything for remote-id `x` if it already has `x` defined (some other value).
- Try to check if I can filter the fetchables by non-use-restricted (to minimize catch for `test?` extra tarballs)
- Decide how to work with mirrors (like `mirror://pypi`)? Maybe some extra match rules?

  On the same note, maybe a new check to catch when using straight pypi hosted files instead of shorter `mirror://pypi`?

PR #486, opened 2022-10-28
MissingRemoteIdCheck: check for missing remote-id

arthurzam:
Scans HOMEPAGE and SRC_URI for uris matching regexes using which it extracts remote-id. Skips already defined remote-id types. Skips URIS that end with ".diff" or ".patch". Prefers to take remote-id from newer package versions, in case URL updated.

Resolves: https://github.com/pkgcore/pkgcheck/issues/475

### To-Do:

- [x] skip fetchables behind use flags
- [x] add more URIs
- [x] add tests


arthurzam:
I think this is good enough set of regexes for inferring. The results are [MissingRemoteIdCheck.log](https://github.com/pkgcore/pkgcheck/files/9893678/MissingRemoteIdCheck.log), which is 1998 suggestions (this was the reason I stopped adding more regexes, as this is exactly my birth year 😄).

Now, we could even automate the addition of those remote-ids to gentoo repo, by using `pkgcheck scan -c MissingRemoteIdCheck --reporter JsonStream` and passing into another utility of adding it.

thesamesam (approved):
Fantastic work!

PR #494, opened 2022-11-16
purer setuptools

arthurzam:
As part of https://github.com/pkgcore/snakeoil/issues/77 , use purer setuptools setup. Some of the big changes:

1. Wheel are python ABI agnostic. While platform specific, they don't link/include python stuff (just tree-sitter-bash), so the library is python agnostic, and we can create a "universal" wheel per platform.
2. I use new snakeoil sphinx extension for man/html generation - removes all logic from `doc/conf.py`
3. Add CI workflow to `release.yml` to create platform wheels for `arm64`, `ppc64le`, `s390x`, for both `musl` and `glibc`
4. Add makefile for easier work with the various commands
5. Enable testing directly using `pytest` command, and not using `setup.py test`


pyproject.toml
```diff
@@ -1,12 +1,115 @@
 [build-system]
 requires = [
-	"wheel", "setuptools",
+	"setuptools",
+	"wheel",
```
mgorny, line 4: Do you need the explicit wheel dep?

pyproject.toml
```diff
@@ -1,12 +1,115 @@
 [build-system]
 requires = [
-	"wheel", "setuptools",
+	"setuptools",
```
mgorny, line 3:
```suggestion
	"setuptools >= 61.0.0",
```

for `pyproject.toml` configuration.

mgorny (commented):
Just a cursory look for now.

pyproject.toml
arthurzam: Surprisingly enough, I do indeed use wheel, to mark my wheel as python ABI agnostic
The usage is with the line:
```python
from wheel.bdist_wheel import bdist_wheel as orig_bdist_wheel

...

class bdist_wheel(orig_bdist_wheel):
    ...
```

mgorny (approved):
LGTM

PR #495, opened 2022-11-24
PythonCheck: updates for py2 removal

mgorny:
Basically:
- clean up after old packages (`dev-python/pypy*-bin`, `virtual/pypy*`)
- stop warning about deps on `dev-python/pypy` or `dev-lang/python:2.7`
- update tests to use py3 targets throughout


mgorny:
FTR, I've left reporting `=dev-lang/python-2*` as-is as that is a corner case that I don't it's worth supporting.

arthurzam:
@mgorny I pushed a commit to the branch, which fixes the test fails - please amend it with correct commit from the list (no need for credit) :)

arthurzam (approved):
Other than the amend, all changes looks great - thank you!

mgorny:
Ah, sorry. I had it locally but must've forgotten to push. Merged your version now, thanks!

PR #500, opened 2022-12-10
MissingRemoteIdCheck: give ready <remote-id/> sample

mgorny:
Instead of verbose description of what to put in the missing remote-id element, just print a ready-to-use XML snippet.

Signed-off-by: Michał Górny <mgorny@gentoo.org>


arthurzam:
Umm, I like the idea but this is not like other descriptions.

@thesamesam - do you think this would be improvement?

Issue #502, opened 2022-12-27
Traceback on `pkgdev push` (AttributeError: module 'pkgcore.ebuild.atom' has no attribute 'valid_use_flag')


thesamesam:
When running `pkgdev push` this morning with latest pkgcore stack (9999 ebuilds):
```
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/pkgcheck/addons/__init__.py", line 288, in init_addon
    addon = addons_map[cls]
            ~~~~~~~~~~^^^^^
KeyError: <class 'pkgcheck.checks.metadata.IuseCheck'>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python-exec/python3.11/pkgdev", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/lib/python3.11/site-packages/pkgdev/scripts/__init__.py", line 46, in main
    run(os.path.basename(sys.argv[0]))
  File "/usr/lib/python3.11/site-packages/pkgdev/scripts/__init__.py", line 38, in run
    sys.exit(tool())
             ^^^^^^
  File "/usr/lib/python3.11/site-packages/snakeoil/cli/tool.py", line 81, in __call__
    ret = self.main()
          ^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/pkgdev/cli.py", line 24, in main
    return super().main()
           ^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/snakeoil/cli/tool.py", line 191, in main
    self.handle_exec_exception(e)
  File "/usr/lib/python3.11/site-packages/snakeoil/cli/tool.py", line 176, in main
    exitstatus = func(self.options, self.out, self.err)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/pkgdev/scripts/pkgdev_push.py", line 55, in _push
    pipe = scan(options.scan_args)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/pkgcheck/api.py", line 44, in scan
    return Pipeline(options)
           ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/pkgcheck/pipeline.py", line 36, in __init__
    self._pipes = self._create_runners()
                  ^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #503, opened 2022-12-27
Fix API compatibility w/ pkgcore revision eb6f4edd2

ferringb:
In that revision, USE flag validation is moved to eapi objects; EAPI ultimately arbitrates that, and it cleaned up some internal issues in atom.

However, it broke pkgcheck's metadata check which was using that regex directly; that code's broken anyways- any demandloaded object must not be imported into another scope; the lazy replacement only affects the original scope, thus that `self.valid_use = ` was pinning the proxied regex.

Either way; since EAPI objects now have a use flag validation method, use that.

Fixes pkgcore/pkgcheck#502

Signed-off-by: Brian Harring <ferringb@gmail.com>

### DIFF
--- a/src/pkgcheck/checks/metadata.py
+++ b/src/pkgcheck/checks/metadata.py
@@ -210,11 +210,10 @@ class IuseCheck(Check):
     def __init__(self, *args, use_addon):
         super().__init__(*args)
         self.iuse_handler = use_addon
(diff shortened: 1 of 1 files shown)


ferringb:
How do folks sequence changes to pkgcheck that are intended to reflect pkgcore changes that aren't yet version released?

Issue #504, opened 2022-12-27
Check for redundant inherits via `@PROVIDES`


mgorny:
Right now we're using `@PROVIDES` not to report missing inherits for stuff indirectly exposed via other eclasses. Perhaps it'd be valuable to also report redundant inherits when we're only using the stuff that's exposed by another eclass.

For example, `distutils-r1` provides `python-r1 python-single-r1`. In that case, if `distutils-r1` is inherited, pkgcheck would report direct `python-r1` and `python-single-r1` inherits.

Might be a good idea to check the results first to see if people wouldn't prefer direct inherits in some cases.

arthurzam:
Did a simple implementation for this, here are 253 results:

<details>
  <summary>Results</summary>
  
  ```
  app-admin/ryzen_smu
    ProvidedEclassInherit: version 0.1.2: redundant eclass inherit 'linux-info', provided by 'linux-mod', line 6
    ProvidedEclassInherit: version 0.1.2_p20210627: redundant eclass inherit 'linux-info', provided by 'linux-mod', line 6
    ProvidedEclassInherit: version 0.1.2_p20211205: redundant eclass inherit 'linux-info', provided by 'linux-mod', line 6
  
  app-antivirus/lkrg
    ProvidedEclassInherit: version 0.9.5: redundant eclass inherit 'linux-info', provided by 'linux-mod', line 6
    ProvidedEclassInherit: version 0.9.6: redundant eclass inherit 'linux-info', provided by 'linux-mod', line 6
  
  app-cdr/gcdemu
    ProvidedEclassInherit: version 3.2.6: redundant eclass inherit 'xdg-utils', provided by 'gnome2-utils', line 8
  
  app-crypt/eid-mw
    ProvidedEclassInherit: version 5.1.4-r1: redundant eclass inherit 'xdg-utils', provided by 'gnome2-utils', line 6
    ProvidedEclassInherit: version 9999: redundant eclass inherit 'xdg-utils', provided by 'gnome2-utils', line 6
  
  app-editors/mousepad
    ProvidedEclassInherit: version 0.5.10: redundant eclass inherit 'xdg-utils', provided by 'gnome2-utils', line 6
  
  app-emulation/faudio
    ProvidedEclassInherit: version 21.01: redundant eclass inherit 'multilib-build', provided by 'cmake-multilib', line 10
    ProvidedEclassInherit: version 9999: redundant eclass inherit 'multilib-build', provided by 'cmake-multilib', line 10
[...]

PR #509, opened 2022-12-30
ProvidedEclassInherit: new check for inheriting provided eclases

arthurzam:
Resolves: https://github.com/pkgcore/pkgcheck/issues/504

### DIFF
--- a/src/pkgcheck/checks/eclass.py
+++ b/src/pkgcheck/checks/eclass.py
@@ -98,18 +98,36 @@ class MisplacedEclassVar(results.LineResult, results.Error):
         return f"invalid pre-inherit placement, line {self.lineno}: {self.line!r}"
 
 
+class ProvidedEclassInherit(results.LineResult, results.Style):
+    """Ebuild inherits an eclass which is already provided by another eclass.
+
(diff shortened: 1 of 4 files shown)


src/pkgcheck/checks/eclass.py
```diff
@@ -98,18 +98,38 @@ def desc(self):
...
 
+class ProvidedEclassInherit(results.LineResult, results.Style):
+    """An ebuild inherits eclass which is already provided by another eclass.
+
+    When inheriting an eclass which declares ``@PROVIDES``, those eclasses
+    are guaranteed to be provided by the eclass. Therefore, inheriting them
+    in ebuild is redundant and should be removed.
+    """
+
+    skip_eclasses = ("multibuild", "multibuild-minimal")
```
mgorny, line 109:
```suggestion
    skip_eclasses = ("multibuild",)
```

What's `multibuild-minimal`? xP

That said, I'm still wondering if exclusion list is the right approach. Perhaps we shouldn't be doing `@PROVIDES` for multibuild like this, and instead redocument the vars in the relevant eclasses. After all, they have kind of special values there.

thesamesam (approved):

PR #528, opened 2023-01-19
Use REPO profile base for profile caching

doctaweeks:
Prior to this change, scanning an overlay would always cause a profile cache update since nothing repo-specific was actually getting updated.

Signed-off-by: Daniel M. Weeks <dan@danweeks.net>


arthurzam (approved):
Thank you for the contribution

Issue #526, opened 2023-01-18
Check for arch.list <-> arches.desc missync


mgorny:
If `arches.desc` is present in repo, it should list the same arches as `arch.list`.

As of gentoo/gentoo@141ba52541e9fafb5da1fcf053bbf8a9c124f2a9, this should trigger on ::gentoo because `riscv-linux` is missing from `arches.desc`.

PR #529, opened 2023-01-19
ArchesMisSync: check for arch.list <-> arches.desc missync

arthurzam:
Resolves: https://github.com/pkgcore/pkgcheck/issues/526

### DIFF
--- a/src/pkgcheck/checks/profiles.py
+++ b/src/pkgcheck/checks/profiles.py
@@ -498,6 +498,20 @@ class NonexistentCategories(results.ProfilesResult, results.Warning):
         return f"nonexistent profiles/categories entr{ies}: {categories}"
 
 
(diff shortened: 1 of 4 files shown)


src/pkgcheck/checks/profiles.py
```diff
@@ -614,3 +629,7 @@ def finish(self):
 
         if unused_profile_dirs := available_profile_dirs - seen_profile_dirs:
             yield UnusedProfileDirs(sorted(unused_profile_dirs))
+
+        if arches_desc := frozenset().union(*self.repo.config.arches_desc.values()):
```
mgorny, line 633: Nice trick. I'd have used `itertools.chain()` myself.

mgorny (approved):
LGTM, though sam is probably right about the name.

PR #534, opened 2023-01-29
checks: Add a check for Python pkgnames matching PyPI

mgorny:
Matching the proposed policy. No tests yet.

In any case, let's use it locally and wait for merging it before the things settle a bit and people aren't spammed with "rename everything at once".


src/pkgcheck/checks/python.py
```diff
@@ -694,3 +694,39 @@ def feed(self, pkg):
                 if GITHUB_ARCHIVE_RE.match(uri):
                     yield PythonGHDistfileSuffix(f.filename, uri, pkg=pkg)
                     break
+
+
+class PythonMismatchedPackageName(results.PackageResult, results.Info):
+    """Package name does not following PyPI-based naming policy"""
```
thesamesam, line 700: does not folkow

src/pkgcheck/checks/python.py
```diff
@@ -694,3 +694,39 @@ def feed(self, pkg):
...
+
+    def __init__(self, expected, **kwargs):
```
arthurzam, line 702: Slowly I'm trying to include typing where possible

```suggestion
    def __init__(self, expected: str, **kwargs):
```

arthurzam (changes requested):
Let's also add a tests (should be simple) for:

1. single matching
2. single not-matching
3. single matching with `.`->`-` rule
4. multiple all not-matching

mgorny:
Added some typing and relaxed the rule to require "normalized name" match.

PR #535, opened 2023-01-29
skip annotations for version results

### LINKED ISSUE #478: Support annotations to silence warnings in ebuilds
I thought we had a bug for this, huh.

We need a way to annotate ebuilds to disable warnings where there's either a policy exemption for something (think e.g. toolchain ebuilds which often do something gnarly), or the warning is a false positive (see https://github.com/pkgcore/pkgcheck/issues/359 for `TarballAvailable`, https://github.com/pkgcore/pkgcheck/issues/214 for `MissingUnpackerDep`).

In previous discussions, we've agreed it'd be best to have it be ebuild-wide (so any annotations would affect the whole ebuild) instead of trying to do it per-line (given it's hard to then restrict for all checks, it'd be ugly if you had to place it on a specific line above or so, and it's not worth the complexity).

arthurzam:
Add support for an ebuild to declare a skip annotation for a class of **version** results. This skip annotation must be declared before any non-comment non-empty line (most of the times it would be before the EAPI declaration). Empty lines between the copyright and the annotation are allowed.
The annotation line must be precise, only one class per line and every space is required.

For example, it would look like:
```bash
# Copyright 2023 Gentoo Authors
# Distributed under the terms of the GNU General Public License v2

# pkgcheck skip: PythonCompatUpdate

# Approved by larry 2023-04-31
# pkgcheck skip: VariableScope

EAPI=8
```

Resolves: https://github.com/pkgcore/pkgcheck/issues/478

## More thoughts

1. Is the format fine?
2. Before merge to master/before release, we must define a QA policy who and when can add a skip.
   As preliminary idea from me (non QA member!): error none, warning only with QA approval, info/style anyone?
3. Am I too hard with the syntax and placing?
4. How and if we should support annotation for package ("dir") results?
5. Should I support multiple class names in same line, separated with commas?
6. Should I add special syntax for placing QA member name and date for those that require approval?
7. Should I post to -dev ML when ready?

TODO:

- [ ] This should be just after EAPI line instead of before
- [ ] Should support comma separated list of classes on the same line
- [ ] Post to ML for comments
- [ ] Open Question for ML: auto approval on ebuild bump?

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -10,6 +10,18 @@ Contents:
     news
     api
     man/pkgcheck
+    man/pkgcheck/scan
+    man/pkgcheck/keywords
+    man/pkgcheck/checks
+    man/config
+    man/skip
+
+Reporting Bugs
+==============
+
+Please submit an issue via github:
+
+https://github.com/pkgcore/pkgcheck/issues
 
 Indices and tables
 ==================
--- a/doc/man/pkgcheck.rst
+++ b/doc/man/pkgcheck.rst
@@ -12,6 +12,7 @@ pkgcheck
 .. include:: pkgcheck/reporters.rst
 
 .. include:: config.rst
+.. include:: skip.rst
 
 Reporting Bugs
 ==============
--- a/doc/man/skip.rst
+++ b/doc/man/skip.rst
@@ -1,0 +1,28 @@
+Skip results annotations
+========================
+
+pkgcheck supports skipping ebuilds results via special annotation. This is
+useful in cases where you are aware of false positive in pkgcheck check, and
+fixing pkgcheck is hard or impossible.
+
+The annotation must appear in the ebuild before any non-comment non-empty line
+(most of the times it would be just before the EAPI declaration line and after
+the copyright lines). Empty lines between the copyright and the annotation are
+allowed.
+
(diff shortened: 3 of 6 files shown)


radhermit:
If this is going to be supported in the discussed format, I'd highly recommend pulling the pragma values during source object iteration into packages... basically wrap actual package objects adding an attribute for the result class skip set (e.g. see how the bash parse tree source works). This change currently operates at a result level which, in theory, can result in a lot of redundant work (you're reading the ebuild file per result and each ebuild can spawn lots of results), not to mention having to special case package types.

In theory, implementing it like that would even make it possible to add the contextual information into the pipeline before the checks get run so packages that ignore all the results a check can return wouldn't even run the check.

arthurzam:
An update based on benchmarks done by sam, mgorny and me.
The performance impact of the new implementation is negligible when done on full tree.

ferringb:
@arthurzam is this something you still consider worth following up on?

thesamesam:
It's probably one of the most common pkgcheck feature requests - to have a way to silence a warning. I think we want it.

ferringb:
> It's probably one of the most common pkgcheck feature requests - to have a way to silence a warning. I think we want it.

Confirming: this is basically what I laid out in #763?

thesamesam:
Yes.

PR #543, opened 2023-02-07
checks: Add a check for PyPI URL replacement with pypi.eclass

mgorny:
I think it's good enough for the first review round.

General idea:

1. If there's exactly one `SRC_URI` that matches URI and filename exactly, suggest inheriting the eclass and removing `SRC_URI`. Otherwise, suggest replacements for every matched URI.
2. Suggest uses of `pypi_sdist_url` and `pypi_wheel_url` with common normalizations like `"${PN/-/.}"`, and strip arguments that match defaults.
3. If wheel is renamed to `.whl.zip` or `.zip` (with no other changes in filename), suggest using `--unpack` to get roughly the same result. Otherwise, let the user figure out the rename.

<details><summary>Example results</summary>

```
dev-python/installer
  PythonInlinePyPIURI: version 0.6.0: inline PyPI URI found: https://files.pythonhosted.org/packages/py3/i/installer/installer-0.6.0-py3-none-any.whl, inherit pypi.eclass and replace with $(pypi_wheel_url --unpack)

dev-python/zipp
  PythonInlinePyPIURI: version 3.11.0: inline PyPI URI found matching pypi.eclass default, inherit the eclass and remove SRC_URI instead

dev-python/zope-testing
  PythonInlinePyPIURI: version 4.10-r2: inline PyPI URI found: https://files.pythonhosted.org/packages/source/z/zope.testing/zope.testing-4.10.tar.gz, inherit pypi.eclass and replace with $(pypi_sdist_url "${PN/-/.}")

dev-python/wtforms
  PythonInlinePyPIURI: version 3.0.1-r1: inline PyPI URI found: https://files.pythonhosted.org/packages/source/W/WTForms/WTForms-3.0.1.tar.gz, inherit pypi.eclass and replace with $(pypi_sdist_url WTForms)

dev-python/whoosh
  PythonInlinePyPIURI: version 2.7.4-r2: inline PyPI URI found: https://files.pythonhosted.org/packages/source/W/Whoosh/Whoosh-2.7.4.tar.gz, inherit pypi.eclass and replace with $(pypi_sdist_url "${PN^}")

dev-python/termcolor
  PythonInlinePyPIURI: version 2.1.1: inline PyPI URI found: https://files.pythonhosted.org/packages/source/t/termcolor/termcolor-2.1.1.tar.gz, inherit pypi.eclass and replace with $(pypi_sdist_url)
# (there are additional distfiles, i suppose we could improve this to suggest += when there's one pypi URI + other URIs)

dev-python/kombu
  PythonInlinePyPIURI: version 5.3.0_beta2: inline PyPI URI found: https://files.pythonhosted.org/packages/source/k/kombu/kombu-5.3.0b2.tar.gz, inherit pypi.eclass and replace with $(pypi_sdist_url "${PV}" "${PV/_beta/b}")
```
</details>


src/pkgcheck/checks/python.py
```diff
@@ -725,6 +744,101 @@ def feed(self, pkg):
                     yield PythonGHDistfileSuffix(f.filename, uri, pkg=pkg)
                     break
 
+    def simplify_pn_pv(self, pn: str, pv: str, pkg, allow_none: bool
+                       ) -> typing.Tuple[str, str]:
```
arthurzam, line 748: `tuple[str, str]`

src/pkgcheck/checks/python.py
```diff
@@ -725,6 +744,101 @@ def feed(self, pkg):
...
+            expected = f"{PYPI_URI_PREFIX}source/{pkg.package[0]}/{pkg.package}/{expected_fn}"
+            if uris[0] == expected and next(iter(fetchables)).filename == expected_fn:
+                yield PythonInlinePyPIURI(uris[0], pkg=pkg)
+                return
+
+        # otherwise, yield result for every URL, with suggested replacement
+        for f in fetchables:
+            for uri in f.uri:
+                source_match = PYPI_SDIST_URI_RE.match(uri)
+                if source_match is not None:
```
arthurzam, line 801:
```suggestion
                if source_match := PYPI_SDIST_URI_RE.match(uri):
```

src/pkgcheck/checks/python.py
```diff
@@ -725,6 +744,101 @@ def feed(self, pkg):
...
+                    pn, pv, suffix = source_match.groups()
+
+                    if suffix == ".tar.gz":
+                        suffix = None
+                    pn, pv = self.simplify_pn_pv(pn, pv, pkg, suffix is None)
+
+                    args = [arg for arg in ("pypi_sdist_url", pn, pv, suffix) if arg is not None]
```
arthurzam, line 808: I don't know if this considered better? But consider:

`args = tuple(filter(None, ("pypi_sdist_url", pn, pv, suffix)))`

(When filter get's `None` as function, it filters all those that `bool(` return True, in this case as string it will drop `None` and `""`, and `""` is impossible from the function, so looks equivalent).

src/pkgcheck/checks/python.py
```diff
@@ -725,6 +744,101 @@ def feed(self, pkg):
...
+                    yield PythonInlinePyPIURI(uri, args, pkg=pkg)
+                    continue
+
+                wheel_match = PYPI_WHEEL_URI_RE.match(uri)
+                if wheel_match is not None:
```
arthurzam, line 813:
```suggestion
                if wheel_match := PYPI_WHEEL_URI_RE.match(uri):
```

src/pkgcheck/checks/python.py
```diff
@@ -725,6 +744,101 @@ def feed(self, pkg):
...
+                                      f"{pn}-{pv}-{pytag}-{abitag}.zip"):
+                        unpack_arg = "--unpack"
+                    if abitag == "none-any":
+                        abitag = None
+                    if pytag == "py3" and abitag is None:
+                        pytag = None
+                    pn, pv = self.simplify_pn_pv(pn, pv, pkg, abitag is None)
+
+                    args = [arg for arg in ("pypi_wheel_url", unpack_arg, pn, pv, pytag, abitag)
+                            if arg is not None]
```
arthurzam, line 827: similar consideration from above is you think it would be better?

src/pkgcheck/checks/python.py
```diff
@@ -687,29 +695,40 @@ def desc(self):
...
+            return (f"inline PyPI URI found: {self.url}, inherit pypi.eclass and replace with "
+                    f"$({' '.join(self.replacement)})")
+
+
+class PythonFetchableCheck(Check):
+    """Perform Python-specific checks to fetchables."""
 
     required_addons = (addons.UseAddon,)
-    known_results = frozenset([PythonGHDistfileSuffix])
+    known_results = frozenset([PythonGHDistfileSuffix, PythonInlinePyPIURI])
```
arthurzam, line 720: Ongoing process, but please replace with set initializer (takes less time, and I think less bytecode, and should be optimized in py3.11 to just use straight frozenset):
```suggestion
    known_results = frozenset({PythonGHDistfileSuffix, PythonInlinePyPIURI})
```

arthurzam (changes requested):

mgorny:
This is now on hold as I've submitted https://github.com/gentoo/gentoo/pull/29526. Let's get the eclass in the right shape first.

src/pkgcheck/checks/python.py
mgorny: Well, it ends up shorter, so I'm happy with it ;-).

mgorny:
Ok, I've updated the check to match the recent updates to pypi.eclass and I think it's state of art logic right now.

I don't exactly like how many conditions the result class has but I guess we don't want to have two different result classes for the same problem.

<details><summary>Some results</summary>

```
dev-python/Babel
  PythonInlinePyPIURI: version 2.11.0: inline PyPI URI found matching pypi.eclass default, inherit the eclass, set PYPI_NO_NORMALIZE=1, and use SRC_URI+= for other URIs instead

dev-python/APScheduler
  PythonInlinePyPIURI: version 3.9.1: inline PyPI URI found matching pypi.eclass default, inherit the eclass, set PYPI_NO_NORMALIZE=1, and remove SRC_URI instead

dev-python/aiosqlite
  PythonInlinePyPIURI: version 0.18.0: inline PyPI URI found matching pypi.eclass default, inherit the eclass, and remove SRC_URI instead

dev-python/automat
  PythonInlinePyPIURI: version 20.2.0-r1: inline PyPI URI found: https://files.pythonhosted.org/packages/source/A/Automat/Automat-20.2.
0.tar.gz, inherit pypi.eclass and replace with $(pypi_sdist_url --no-normalize "${PN^}")

dev-python/binaryornot
  PythonInlinePyPIURI: version 0.4.4-r2: inline PyPI URI found: https://files.pythonhosted.org/packages/source/b/binaryornot/binaryornot-0.4.4.tar.gz, inherit pypi.eclass and replace with $(pypi_sdist_url)

dev-python/cherrypy
  PythonInlinePyPIURI: version 18.8.0: inline PyPI URI found: https://files.pythonhosted.org/packages/source/C/CherryPy/CherryPy-18.8.0.tar.gz, inherit pypi.eclass and replace with $(pypi_sdist_url --no-normalize CherryPy)

dev-python/collective-checkdocs
  PythonInlinePyPIURI: version 0.2-r1: inline PyPI URI found: https://files.pythonhosted.org/packages/source/c/collective.checkdocs/collective.checkdocs-0.2.zip, inherit pypi.eclass and replace with $(pypi_sdist_url --no-normalize "${PN/-/.}" "${PV}" .zip)

dev-python/cryptography
  PythonInlinePyPIURI: version 38.0.4: inline PyPI URI found: https://files.pythonhosted.org/packages/source/c/cryptography/cryptography-38.0.4.tar.gz, inherit pypi.eclass and replace with $(pypi_sdist_url)
  PythonInlinePyPIURI: version 38.0.4: inline PyPI URI found: https://files.pythonhosted.org/packages/source/c/cryptography_vectors/cryptography_vectors-38.0.4.tar.gz, inherit pypi.eclass and replace with $(pypi_sdist_url cryptography_vectors)

dev-python/pip
  PythonInlinePyPIURI: version 22.3.1: inline PyPI URI found: https://files.pythonhosted.org/packages/py2.py3/w/wheel/wheel-0.36.2-py2.py3-none-any.whl, inherit pypi.eclass and replace with $(pypi_wheel_url wheel 0.36.2 py2.py3)
  PythonInlinePyPIURI: version 22.3.1: inline PyPI URI found: https://files.pythonhosted.org/packages/py3/s/setuptools/setuptools-62.3.2-py3-none-any.whl, inherit pypi.eclass and replace with $(pypi_wheel_url setuptools 62.3.2)

dev-python/selenium
  PythonInlinePyPIURI: version 4.5.0: inline PyPI URI found: https://files.pythonhosted.org/packages/py3/s/selenium/selenium-4.5.0-py3-none-any.whl, inherit pypi.eclass and replace with $(pypi_wheel_url)
```
</details>

I guess I can still improve it to suggest `+=` when there's default sdist + other PyPI URL.

mgorny:
I've added the initial set of tests. Fixes coming up next.

mgorny:
Fixes added. I suppose it'd also be useful to add sth like `python-inline-pypi-url` test package, that tests suggesting `-` substitutions.

mgorny:
Ok, I think it's ready.

arthurzam (approved):

PR #547, opened 2023-02-13
python: Fix project normalization regex to replace runs of [._-]

mgorny:
Fix the regex used for project normalization to replace runs of consecutive `[._-]` symbols with a single hyphen, rather than every     character individually.  Adjust the test to verify that it handles     these sequences correctly.


mgorny:
I've also added other changes preparing for `pypi-eclass` branch.

arthurzam:
Merged

PR #551, opened 2023-02-23
network: add kde-invent remote-id

thesamesam:


thesamesam:
I think this should pass once pkgcore is updated.

Issue #498, opened 2022-11-27
Warn on no-revbump for PEP517


thesamesam:
Enabling PEP517 mode is a non-conservative change and may result in installed files changing. It should always be done in a new revision.

thesamesam:
We need to handle this before too much of the tree is converted to PEP517, as it'll stop being useful then.

arthurzam:
Currently the issue is that in the git checks, we use the not-parsed ebuilds, which results in my inability to know the state of PEP517. This is the reason for it being postponed.

thesamesam:
I think we can do it with reduced accuracy if we regex-match the text (like we do for the copyright check) but I'm not sure how to get the old ebuild text.

arthurzam:
> but I'm not sure how to get the old ebuild text.

https://github.com/pkgcore/pkgcheck/blob/861c24ab76bfc36de2ca80a8375866e4123dc51a/src/pkgcheck/checks/git.py#L364-L376

PR #556, opened 2023-03-04
checks: git: add PythonPEP517WithoutRevbump

thesamesam:
Fixes: https://github.com/pkgcore/pkgcheck/issues/498

TODO:
- [x] Add tests
- [x] Hoist up re.compile

### DIFF
--- a/src/pkgcheck/checks/git.py
+++ b/src/pkgcheck/checks/git.py
@@ -171,6 +171,18 @@ class MissingMove(results.PackageResult, results.Error):
         return f"renamed package: {self.old} -> {self.new}"
 
 
+class PythonPEP517WithoutRevbump(results.PackageResult, results.Warning):
+    """Package has started/stopped using DISTUTILS_USE_PEP517 without revbump.
+
+    The package has started or stopped using DISTUTILS_USE_PEP517 without
+    a new revision. PEP517 affects the files installed by a package
+    and might lead to some files missing.
+
+    """
+
+    desc = "changed DISTUTILS_USE_PEP517 without new revision"
+
+
 class SrcUriChecksumChange(results.PackageResult, results.Error):
     """SRC_URI changing checksum without distfile rename."""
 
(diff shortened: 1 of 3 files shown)


src/pkgcheck/checks/git.py
```diff
@@ -364,6 +378,19 @@ def modified_checks(self, pkgs, added):
         if pkg not in added and old_pkg.rdepend != new_pkg.rdepend:
             yield RdependChange(pkg=new_pkg)
 
+        # It's an existing package, but this doesn't fit into
+        # modified_checks() because we're working on the text.
+        python_pep517_regex = re.compile("^DISTUTILS_USE_PEP517=")
```
thesamesam, line 383: I'll move this out so we don't recompile it every time, obviously.

src/pkgcheck/checks/git.py
```diff
@@ -364,6 +378,17 @@ def modified_checks(self, pkgs, added):
         if pkg not in added and old_pkg.rdepend != new_pkg.rdepend:
             yield RdependChange(pkg=new_pkg)
 
+        python_pep517_regex = re.compile("^DISTUTILS_USE_PEP517=")
+
+        def found_pep517_lines(cmp_pkg):
+            return any(python_pep517_regex.match(line) for line in cmp_pkg.ebuild.text_fileobj())
+
+        found_old_pep517_line = found_pep517_lines(old_pkg)
```
thesamesam, line 386: We can guard this with a distutils-r1 inherit check.

arthurzam (approved):

mgorny:
On a semi-relevant note: I wonder if we should have something similar for EAPI bumps.

thesamesam:
> On a semi-relevant note: I wonder if we should have something similar for EAPI bumps.

My view is that we should, for two reasons:
1. Risk of image changing (not just because of EAPI behaviour (also think of Portage keepdir w/ EAPI 8+, but also eclass functions may change)
2. Updating the EAPI in vdb so we can eventually remove old version support from Portage

arthurzam (approved):
Thank you!

thesamesam:
Thanks!

PR #573, opened 2023-04-24
Fix config loading if XDG_CONFIG_HOME is defined

agireud:
Similar issue to https://github.com/pkgcore/pkgdev/issues/72.  Applies a similar fix as https://github.com/pkgcore/pkgdev/commit/9f60576753da4cbe5cd0afe660d49a22e34e3cf1 to match what is already documented in the man page.


arthurzam:
@agireud After another check, you haven't signoff you commit. Please signoff it (regular gentoo GLEP, same as for normal packages)

Issue #581, opened 2023-05-18
more fix required for #576


ZeroChaos-:
strace showing different files are read depending on which directory pkgcheck is run from.  The one I'm specifically expecting it to read is /home/zero/development/pentoo/pentoo-overlay/metadata/pkgcheck.conf
```
zero@theprophet pentoo-overlay % strace pkgcheck scan -C PentooCI 2>&1 | grep -i pkgcheck.conf                                                         (git)-[master]  130 ⨯
openat(AT_FDCWD, "/etc/python-exec/pkgcheck.conf", O_RDONLY) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/usr/share/pkgcheck/pkgcheck.conf", {st_mode=S_IFREG|0644, st_size=717, ...}, 0) = 0
openat(AT_FDCWD, "/usr/share/pkgcheck/pkgcheck.conf", O_RDONLY|O_CLOEXEC) = 3
newfstatat(AT_FDCWD, "/etc/pkgcheck/pkgcheck.conf", 0x7ffdc45e4228, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/home/zero/.config/pkgcheck/pkgcheck.conf", 0x7ffdc45e4228, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/home/zero/development/pentoo/pentoo-overlay/metadata/pkgcheck.conf", {st_mode=S_IFREG|0644, st_size=721, ...}, 0) = 0
openat(AT_FDCWD, "/usr/share/pkgcheck/pkgcheck.conf", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/home/zero/development/pentoo/pentoo-overlay/metadata/pkgcheck.conf", O_RDONLY|O_CLOEXEC) = 3
^C
zero@theprophet pentoo-overlay % cd ..                                                                                                                 (git)-[master]  130 ⨯
zero@theprophet pentoo % strace pkgcheck scan -C PentooCI 2>&1 | grep -i pkgcheck.conf
openat(AT_FDCWD, "/etc/python-exec/pkgcheck.conf", O_RDONLY) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/usr/share/pkgcheck/pkgcheck.conf", {st_mode=S_IFREG|0644, st_size=717, ...}, 0) = 0
[...]

PR #582, opened 2023-05-19
docs/config: explain when repo config is used

arthurzam:
Resolves: https://github.com/pkgcore/pkgcheck/issues/581


ZeroChaos- (approved):
this makes sense to me, is clear, and isn't too verbose.  I like it.

Issue #586, opened 2023-06-18
cargo.eclass checks: detect unnecessary `$(cargo_crate_uris)` and `CRATES` with `-` separator


mgorny:
Given the new optimizations, pkgcheck could report:

1. `CRATES` using `-` instead of `@`.
2. `$(cargo_crate_uris)` and `$(cargo_crate_uris ${CRATES})` being called directly, instead of `${CARGO_CRATE_URIS}`. Note that calling it with different arguments is still valid.

mgorny:
CC @gentoo/rust

mgorny:
CC @ionenwks, @thesamesam

arthurzam:
@mgorny It would help me if you could give some more info, as in the [wiki](https://wiki.gentoo.org/wiki/Project:PkgCore#New_QA_check) (Yes, I know I should create GitHub issues templates for those, TODO...)

I'm not that familiar with rust ebuilds, so having those will make it easier for me :)

mgorny:
Roughly:
- https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=0e0a59b931d9d079aca1347015c122bdc2fb7f58
- https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=21595fc8da27066eb90b11c2c6e24c7a43015641

1. If `CRATES` contains at least one token that doesn't contain `@`, output a warning: `using - as name-version separator in CRATES is suboptimal, use name@version instead`. Detailed help may suggest using pycargoebuild 0.7+ to generate new CRATES.
2. If `$(cargo_crate_uris)`, `$(cargo_crate_uris ${CRATES})` or another variant of the latter with/out quoting and with/out `{}`s is found, output a warning: calling `$(cargo_crate_uris) is suboptimal, use ${CARGO_CRATE_URIS} for global CRATES instead`.

PR #589, opened 2023-06-21
RustCheck: new crates checks

arthurzam:
Resolves: https://github.com/pkgcore/pkgcheck/issues/586

### DIFF
--- a/src/pkgcheck/checks/rust.py
+++ b/src/pkgcheck/checks/rust.py
@@ -17,6 +17,20 @@ class SuboptimalCratesSeparator(results.LineResult, results.Warning):
         return f"line: {self.lineno}: using - as name-version separator in CRATES is suboptimal, use name@version instead"
 
 
+class SuboptimalCratesURICall(results.LineResult, results.Warning):
+    """Calling ``cargo_crate_uris`` with ``CRATES`` is suboptimal, use
+    ``${CARGO_CRATE_URIS}``.
+
+    Calls to ``$(cargo_crate_uris)`` and ``$(cargo_crate_uris ${CRATES})`` are
+    suboptimal, and can be replaces with ``${CARGO_CRATE_URIS}`` which is
+    pre-computed, faster and doesn't require sub-shell in global-scope.
+    """
+
+    @property
+    def desc(self):
+        return f"line: {self.lineno}: calling {self.line!r} is suboptimal, use '${{CARGO_CRATE_URIS}}' for global CRATES instead"
+
+
 class RustCheck(Check):
     """Checks for rust related issues."""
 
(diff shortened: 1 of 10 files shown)


src/pkgcheck/checks/rust.py
```diff
@@ -0,0 +1,47 @@
...
+    def feed(self, pkg: bash.ParseTree):
+        if "cargo" not in pkg.inherited:
+            return
+        for node in pkg.global_query(bash.var_assign_query):
+            name = pkg.node_str(node.child_by_field_name("name"))
+            if name == "CRATES":
+                val_node = node.children[-1]
+                row, _ = val_node.start_point
+                val_str: str = pkg.node_str(val_node).strip("'\"")
+                for lineno, line in enumerate(val_str.splitlines(), start=row + 1):
```
mgorny, line 54: This variable is just whitespace separated, so technically speaking there could be multiple tokens per line.

src/pkgcheck/checks/rust.py
```diff
@@ -35,7 +50,7 @@ def feed(self, pkg: bash.ParseTree):
             if name == "CRATES":
                 val_node = node.children[-1]
                 row, _ = val_node.start_point
-                val_str: str = pkg.node_str(val_node).strip("'\"")
```
mgorny: This change seems to be on the wrong commit.

src/pkgcheck/checks/rust.py
```diff
@@ -45,3 +60,21 @@ def feed(self, pkg: bash.ParseTree):
...
+            call_name = pkg.node_str(node.child_by_field_name("name"))
+            if call_name == "cargo_crate_uris":
+                row, _ = node.start_point
+                line = pkg.node_str(node.parent)
+                if node.child_count == 1 or (
+                    node.child_count == 2
+                    and any(
+                        pkg.node_str(var_node) == "CRATES"
+                        for var_node, _ in bash.var_query.captures(node.children[1])
+                    )
```
mgorny, line 73: I'm not sure if I understand this correctly but the replacement works only if `${CRATES}` (or `"${CRATES}"`) is the only argument.

arthurzam: Yes, that is why I verify the node count is exactly 2. I've even added a test for `${CRATES} something` and verify it ignores it :)

src/pkgcheck/checks/rust.py
arthurzam: True, maybe just use split... will see

arthurzam: fixed

src/pkgcheck/checks/rust.py
mgorny: Well, I guess I don't understand what's the iteration for but I guess it's high magic.

Issue #588, opened 2023-06-21
UnknownProfilePackage: use blocker cache to mention "last match removed"


mgorny:
```
profiles
  UnknownProfilePackage: 'package.mask': unknown package: '<app-admin/eselect-1.4.22-r1'
  UnknownProfilePackage: 'package.mask': unknown package: '=dev-db/mariadb-10.11.2'
  UnknownProfilePackage: 'package.mask': unknown package: '=dev-db/mariadb-10.11.3'
  UnknownProfilePackage: 'package.mask': unknown package: '=dev-db/mariadb-10.5.20'
  UnknownProfilePackage: 'package.mask': unknown package: '=dev-db/mariadb-10.6.13'
...
```

I think it would make sense to tell when was the package removed, if it matched at all.

mgorny:
Thanks!

PR #590, opened 2023-06-21
ProfilesCheck: for unknown packages, mention last match removed

arthurzam:
https://github.com/pkgcore/pkgcheck/issues/588

### DIFF
--- a/src/pkgcheck/checks/profiles.py
+++ b/src/pkgcheck/checks/profiles.py
@@ -1,5 +1,6 @@
 """Various profile-related checks."""
 
+from datetime import datetime
 import os
 from collections import defaultdict
 from typing import Iterable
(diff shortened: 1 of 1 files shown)


src/pkgcheck/checks/profiles.py
```diff
@@ -229,6 +254,17 @@ def __init__(self, *args, use_addon: addons.UseAddon, keywords_addon: addons.Key
             | use_addon.global_iuse_implicit
         )
 
+    def _report_unknown_atom(self, path, atom):
+        if not isinstance(atom, atom_cls):
+            atom = atom_cls(atom)
+        if matches := self.existence_repo.match(atom):
+            removal = max(x.time for x in matches)
+            removal = datetime.fromtimestamp(removal)
+            years = (self.today - removal).days / 365
```
mgorny, line 263: 365 is not a very accurate approximation xP.

arthurzam: It was good enough for `OutdatedBlocker` to not notice. But in true sense, it results in ~4 days since starts of Gentoo's life, which under rounding error is 0.01 of the year. I think we can slide it unless better solution found.

mgorny: A better solution was known for at least a few hundred years ;-).
```suggestion
            years = (self.today - removal).days / 365.2425
```

thesamesam (approved):
Looks reasonable with tests added. Thank you!

Issue #585, opened 2023-06-11
pkgcheck doesn't seem to consider distfile named ".gh.tar.gz" suspicious ;-)


mgorny:
I suppose we should throw an error for any distfile whose name starts with `.` (it's usually a typo like `${MYP}.gh.tar.gz`).

PR #592, opened 2023-06-24
SrcUriFilenameDotPrefix: new check for SRC_URI filenames with dot prefix

arthurzam:
Resolves: https://github.com/pkgcore/pkgcheck/issues/585


thesamesam (approved):
Nice and straightforward. Thanks!

mgorny:
I wonder if we shouldn't simply put it in `bad_filenames`.

arthurzam:
> I wonder if we shouldn't simply put it in `bad_filenames`.

@mgorny I'm unsure about it. While simpler, the BadFilename is quite ambiguous, and I'm unsure anybody will understand what it means in practice. Here we have a more specialized case, where we warn the develop - "you might have done big mistake, reverify".

I think it will be better to have more info than less here.

src/pkgcheck/checks/metadata.py
```diff
@@ -1315,6 +1315,23 @@ class InvalidSrcUri(results.MetadataError, results.VersionResult):
     attr = "fetchables"
 
 
+class SrcUriFilenameDotPrefix(results.VersionResult, results.Warning):
```
mgorny, line 1318:
```suggestion
class SrcUriFilenameDotPrefix(results.VersionResult, results.Error):
```

mgorny (commented):
Yeah, makes sense — but let's make this one fatal though.

src/pkgcheck/checks/metadata.py
arthurzam: Good idea, done

mgorny (approved):
LGTM

Issue #304, opened 2021-03-11
PythonCompatUpdate for lua and ruby, too


juippis:
This just popped to my mind and I wanted to let it out somewhere and you were not in IRC.

I didn't see duplicate for this, nor it being documented in https://pkgcore.github.io/pkgcheck/man/pkgcheck.html yet.

arthurzam:
While I agree with the idea, and it is quite simple to add, I have asked people of lua & ruby projects if they would like that, and in multiple times they rejected it.

juippis:
Thanks for trying!

thesamesam:
I don't know if it makes sense for Lua because Lua often requires proper porting between versions, but for Python and Ruby, it's _largely_ a function of whether one's deps are ready + usually a small amount of porting work in the package itself, if any.

I think the check at least is well-defined for Ruby, even if we don't enable it by default. I'd find this pretty useful over having to do what we used to for Python - manually adding to PYTHON_COMPAT, then running `pkgcheck scan` to look for dep errors.

PR #595, opened 2023-06-29
RubyCompatCheck: new check for new USE_RUBY compatible values

arthurzam:
I think the `ruby-ng.eclass` is a real abomination I've created. I've extracted parts from real ruby-ng, modified as I needed, took shortcuts and more. I hope no sane person will use this abomination, we need to cast in into the fire :)

Resolves: https://github.com/pkgcore/pkgcheck/issues/304

### DIFF
--- a/src/pkgcheck/checks/ruby.py
+++ b/src/pkgcheck/checks/ruby.py
@@ -1,0 +1,109 @@
+import itertools
+
+from pkgcore.ebuild.atom import atom
+from snakeoil.sequences import iflatten_instance
+from snakeoil.strings import pluralism
+
+from .. import results
+from . import Check
+
+
+IUSE_PREFIX = "ruby_targets_"
+
+
+class RubyCompatUpdate(results.VersionResult, results.Info):
+    """``USE_RUBY`` can be updated to support newer ruby version(s)."""
+
+    def __init__(self, updates, **kwargs):
+        super().__init__(**kwargs)
+        self.updates = tuple(updates)
+
+    @property
+    def desc(self):
+        s = pluralism(self.updates)
+        updates = ", ".join(self.updates)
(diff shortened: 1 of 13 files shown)


src/pkgcheck/checks/ruby.py
```diff
@@ -0,0 +1,107 @@
...
+
+from .. import results
+from . import Check
+
+
+IUSE_PREFIX = "ruby_targets_"
+
+
+class RubyCompatUpdate(results.VersionResult, results.Info):
+    """``USE_RUBY`` can be updated to support newer python version(s)."""
```
thesamesam, line 15:
```suggestion
    """``USE_RUBY`` can be updated to support newer Ruby version(s)."""
```

src/pkgcheck/checks/ruby.py
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+    def feed(self, pkg):
+        if "ruby-ng" not in pkg.inherited:
+            return
+
+        deps = self.deps(pkg)
+
+        try:
+            # determine the latest supported python version
```
thesamesam, line 74:
```suggestion
            # determine the latest supported ruby version
```

src/pkgcheck/checks/ruby.py
```diff
@@ -0,0 +1,107 @@
...
+            latest_target = sorted(
+                (
+                    f"ruby{x.slot.replace('.', '')}"
+                    for x in deps
+                    if x.key == "dev-lang/ruby" and x.slot is not None
```
thesamesam, line 79: we should probably ignore virtual/rubygems and virtual/ruby-ssl too

arthurzam: done

thesamesam (approved):

Issue #593, opened 2023-06-24
Feature request, new check: Perl virtual dependencies


akhuettel:
Here's a wish for a new check. (@thesamesam @arthurzam)

- If PN is dev-lang/perl or perl-core/* or virtual/perl-*, skip the check
- Otherwise, all dependency declarations that specify a Perl virtual (virtual/perl-*) and do **not** give any type of version dependency trigger a warning.

Example: [dev-perl/XXX-0.380.0](https://github.com/gentoo/gentoo/blob/66ebb42027c1b2cd505e648ffc90f1bcac8f2855/dev-perl/XXX/XXX-0.380.0.ebuild#L17-L19)
```
BDEPEND="${RDEPEND}
        virtual/perl-ExtUtils-MakeMaker
"
```
Triggers a warning since the dependency on virtual/perl-ExtUtils-MakeMaker in BDEPEND does not specify any version (i.e., no =, <, >, ~, ...)

If this ever goes into CI, all current warnings will have to be grandfathered in since otherwise things explode.

thesamesam:
We can use the implementation of `MissingPackageRevision` for some inspiration, maybe

arthurzam:
1. Is it relevant only to ebuilds inheriting perl eclass, or all packages?
2. Thank for the precise logic, but what is the issue that requires this? I need this for the documentation and explanation :)

akhuettel:
>     1. Is it relevant only to ebuilds inheriting perl eclass, or all packages?

All packages

>     2. Thank for the precise logic, but what is the issue that requires this? I need this for the documentation and explanation :)

The virtuals stand for packages that have releases both as part of dev-lang/perl and standalone. Apart from rare special cases, if you require "any" version of such a virtual, this will always be fulfilled by dev-lang/perl.

For a long time we had the policy that also version-less dependencies on Perl virtuals should be added to packages. I want to change that, since these dependencies are pointless and obscure which virtuals are still needed. Reducing the number of Perl virtuals would be great, and once we have only version dependencies in the tree it's trivial to see which ones are obsolete and which ones not.

PR #597, opened 2023-07-02
PerlCheck: check versioned virtual perl dependencies

arthurzam:
@akhuettel @thesamesam So here it is. A 
[huge amount of matches](https://github.com/pkgcore/pkgcheck/files/11931201/3MHLRYTHV.txt) across the repo. After some talks on IRC (#-perl) this is done as optional check, so folks need to enable `PerlCheck` for this result to appear.

I'm fine with any decision (merge or decline), so don't feel obligated 😄 

Resolves: https://github.com/pkgcore/pkgcheck/issues/593

### DIFF
--- a/src/pkgcheck/checks/perl.py
+++ b/src/pkgcheck/checks/perl.py
@@ -2,8 +2,11 @@ import multiprocessing
 import re
 import subprocess
 
+from pkgcore.ebuild.atom import atom as atom_cls
 from pkgcore.restrictions import packages, values
+from pkgcore.package.errors import MetadataException
 from snakeoil.osutils import pjoin
+from snakeoil.sequences import iflatten_instance
 
 from .. import const, results, sources
 from . import OptionalCheck, SkipCheck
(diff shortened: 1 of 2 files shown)


arthurzam:
So we have:

```
[23:18:57] <@dilfridge> off by default is ok
```

So I guess we can continue with this and review before merge...

src/pkgcheck/checks/perl.py
```diff
@@ -22,6 +25,24 @@ def desc(self):
         return f"DIST_VERSION={self.dist_version} normalizes to {self.normalized}"
 
 
+class MissingOPVirtualPerlDependency(results.VersionResult, results.Warning):
```
thesamesam, line 28: Maybe:
```suggestion
class MissingVersionedVirtualPerlDependency(results.VersionResult, results.Warning):
```

although it's a bit long?

thesamesam (commented):
Thank you Arthur!

Issue #600, opened 2023-07-09
Git remote currently required to be named "origin", please allow other remote names via configuration


hartwork:
Hi!

Currently, the code at https://github.com/pkgcore/pkgcheck/blob/master/src/pkgcheck/addons/git.py hardcodes remote name `origin` in multiple places which e.g. results in weird errors (see below) for users that avoid name `origin` for safety reasons, e.g. my remote is named `git-gentoo-org` instead. 

The symptom with `pkgcheck scan` is this:

```console
# pkgcheck scan --commits 331a06d2a333b02996e9cc49ab55441c9bd749bd^..331a06d2a333b02996e9cc49ab55441c9bd749bd
pkgcheck scan: error: failed retrieving commit hash for git repo: '[..]/gentoo-GIT'
```

Renaming the remote `git remote rename git-gentoo-org origin` works around the issue but forces magic name `origin` on me that I use nowhere else.

I understand if this is not considered high priority.
Thanks for your consideration :pray: 

Sebastian

CC @thesamesam

PR #601, opened 2023-07-09
scan: add `--git-remote` option to select remote

arthurzam:
For repos with multiple remotes, it might be useful to select a specific remote to use (and not the default origin). This can be set using the `--git-remote` option for cmd call, or by adding `git-remote=value` to the config file.

@hartwork mind applying the patch and verifying if it work for you (ideally both ways of cmd arg and [config file](https://pkgcore.github.io/pkgcheck/man/pkgcheck.html#config-file-support).

Resolves: https://github.com/pkgcore/pkgcheck/issues/600

### DIFF
--- a/src/pkgcheck/addons/git.py
+++ b/src/pkgcheck/addons/git.py
@@ -342,15 +342,15 @@ class _ScanGit(argparse.Action):
     def __init__(self, *args, staged=False, **kwargs):
         super().__init__(*args, **kwargs)
         if staged:
-            default_ref = "HEAD"
             diff_cmd = ["git", "diff-index", "--name-only", "--cached", "-z"]
         else:
-            default_ref = "origin..HEAD"
             diff_cmd = ["git", "diff-tree", "-r", "--name-only", "-z"]
 
         self.staged = staged
-        self.default_ref = default_ref
         self.diff_cmd = diff_cmd
+
+    def default_ref(self, remote):
+        return "HEAD" if self.staged else f"{remote}..HEAD"
 
     def generate_restrictions(self, parser, namespace, ref):
         """Generate restrictions for a given diff command."""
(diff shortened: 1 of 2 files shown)


hartwork (commented):
Hi @arthurzam very nice and quick — thank you! :pray: 

I have tried both (1) `--git-remote git-gentoo-org` and (2) the config option with no `origin` in sight now and it worked with each of these in isolation and did not work by default (as expected). That's great.

For anyone else interested, the config file looked like this for me in practice:
```
# cat ~/.config/pkgcheck/pkgcheck.conf
[DEFAULT]
git-remote = git-gentoo-org
```

One thing I noticed is that in the config file `git-remote` (with a dash) could be an unusual key format/syntax here if nothing else uses a dash so far (not sure); if so, `git_remote` with an underscore may work better, just an idea.

arthurzam:
> Hi @arthurzam very nice and quick — thank you! pray
> 
> I have tried both (1) `--git-remote git-gentoo-org` and (2) the config option with no `origin` in sight now and it worked with each of these in isolation and did not work by default (as expected). That's great.

Thank you :)

> For anyone else interested, the config file looked like this for me in practice:
> 
> ```
> # cat ~/.config/pkgcheck/pkgcheck.conf
> [DEFAULT]
> git-remote = git-gentoo-org
> ```
> 
> One thing I noticed is that in the config file `git-remote` (with a dash) could be an unusual key format/syntax here if nothing else uses a dash so far (not sure); if so, `git_remote` with an underscore may work better, just an idea.

At pkgcore stack we use the `-` symbol for command line arguments word separation (python's argparse converts it into `_` for the option variable name). I think this style also makes better sense compared to `git` and the `--` prefix :) so we should be same style. But thank you for the concern and verifying it.

I'll merge it now.

PR #603, opened 2023-07-15
[wip] RDEPEND cycle check

### LINKED ISSUE #400: Detect unconditional RDEPEND-RDEPEND cycles
Unfortunately, since Portage maintainers have historically preferred to add ugly hacks over fixing ebuilds, Portage currently pretends that "pure RDEPEND" cycles aren't there. It would be nice if pkgcheck tries to detect at least unconditional cycles, e.g. A unconditionally RDEP on B, B on C, C on A, and requested developers to use PDEPEND in order to achieve predictable merge order.

arthurzam:
This is WIP, showing results across the repo, but with possible false positives (since we compare on `cat/pkg` level instead of `cat/pkg-ver` levels).

Resolves: https://github.com/pkgcore/pkgcheck/issues/400

### DIFF
--- a/src/pkgcheck/checks/visibility.py
+++ b/src/pkgcheck/checks/visibility.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+from itertools import groupby
 from operator import attrgetter
 
 from pkgcore.ebuild.atom import atom, transitive_use_atom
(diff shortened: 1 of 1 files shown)


src/pkgcheck/checks/visibility.py
```diff
@@ -386,3 +386,74 @@ def process_depset(self, pkg, attr, depset, edepset, profiles):
...
+        pkg_deps = {
+            pkg: {dep.key for dep in pkg.rdepend if isinstance(dep, atom) and not dep.blocks}
+            for pkg in pkgset
+        }
+        self.visited_packages[key] = all_deps = frozenset().union(*pkg_deps.values())
+        if missing := all_deps - self.visited_packages.keys():
+            for missing_key in missing:
+                try:
+                    self._collect_deps_graph(self.repo.match(atom(missing_key)))
+                except IndexError:
```
thesamesam, line 441: Let's put a comment for how it can happen (nonexistent/invalid dep, `NonexistentDeps`)

thesamesam:
net-libs/gssdp has a FP: gssdp-1.6.2 depends on gssdp-1.4:0, but it thinks there's a cycle

not checked others yet

thesamesam (commented):
It looks good so far, but not tried it yet.

PR #611, opened 2023-08-05
RepoDirCheck: use charset_normalizer.is_binary()

cjmayo:
Saves maintaining some code here, dev-python/requests already requires charset-normalizer and installing chardet means requests (and bs4) will use that instead.
But I've only updated some of the tests (I can't see what is failing the others) and charset-normalizer-3.2.0 is new and not yet stabilised in Gentoo.

FAILED tests/scripts/test_pkgcheck_scan.py::TestPkgcheckScan::test_scan_verify[gentoo] - Failed: unmatched repo scan results:
FAILED tests/scripts/test_pkgcheck_scan.py::TestPkgcheckScan::test_fix[RepoDirCheck-BinaryFile] - Failed: unexpected repo scan results:


arthurzam (changes requested):
Sadly the failures are looking severe, so we need to rework it to work. I'm not against switching to `charset-normalizer` (doesn't add a big dep tree behind it).

src/pkgcheck/checks/repo.py
```diff
@@ -1,10 +1,10 @@
 import os
 
+from charset_normalizer import is_binary
```
arthurzam, line 3: The problem with the tests is that the selected file has content of `..\n`, which is decided as `is_binary = True`, which isn't correct.

Issue #622, opened 2023-09-15
[New Check]: Multiple assignments to a variable (maybe just eclass vars?) in global scope


thesamesam:
### Is there an existing such new check request for this?

- [X] I have searched the existing issues

### Explain

Assigning twice to a variable in global scope is a serious code smell and is likely a typo:
```bash
SRC_URI="https://example.com/foo.tar.gz"
S="${WORKDIR}/x"

LICENSE="..."

S="${WORKDIR}"
src_compile() {
:;
}

```

We could also restrict the check to just eclass variables, although I'd like to try it without that restriction first.

C more or less calls this `-Wshadow` (not quite as there's some interaction w/ scoping which bash lacks).

Examples:
* Double-assignment to `S` (perhaps once under `SRC_URI` and another way below above phases, might not be easy to see in a long ebuild)
* Setting `PYTHON_COMPAT` twice in an ebuild

### Examples

pecl-pttp (https://github.com/gentoo/gentoo/pull/32762/files#diff-c95f1d4880dbd4fd60024792709466e4113c4a06fe6d46b5b660c00e55f5440a, https://github.com/gentoo/gentoo/blob/4463a8bd59a49bdba73880461f3cedccf6b8cdb6/dev-php/pecl-http/pecl-http-3.3.0.ebuild#L16C5-L16C5) has `USE_PHP` defined twice

### Output message

VariableShadowed

### Documentation

Variable is shadowed / repeatedly declared. This is a possible typo.

### Result level

warning

thesamesam:
There's a chance that we need to restrict this a bit, either to:
* only-PMS vars, or
* exclude eclass vars, or
* exclude lines with a function call, e.g. `USE_PHP=... php_blah` (my preference if we have to restrict it at all)

The latter two would be to cover cases where you set something like `USE_RUBY` globally but you need a hack to restrict the dependencies added by a function, like:
```
USE_RUBY="ruby31 ruby32"
inherit ruby-ng
# ...
USE_RUBY="ruby31" ruby_add_bdepend "dev-ruby/foo" # only needed for ruby31, as it's built in to ruby32 (think of this like a hack for Python's gen_cond_dep)
```

PR #623, opened 2023-09-15
DeclarationShadowedCheck: new checks to detect shadowed variables & functions

arthurzam:
Rules:

- Defined in global scope
- Not inside logic block (so only check top level assignments)
- Doesn't use variable append (`BDEPEND+=""`)
- Doesn't mention self variable inside (`BDEPEND="${BDEPEND} ..."`)

Resolves: https://github.com/pkgcore/pkgcheck/issues/622

- [x] Add tests

### DIFF
--- a/src/pkgcheck/checks/codingstyle.py
+++ b/src/pkgcheck/checks/codingstyle.py
@@ -1443,3 +1443,43 @@ class GlobCheck(Check):
                         lineno, _colno = node.start_point
                         yield GlobDistdir(line=pkg.node_str(node), lineno=lineno + 1, pkg=pkg)
                         break
+
+
+class VariableShadowed(results.LinesResult, results.Warning):
+    """Variable is shadowed or repeatedly declared. This is a possible typo."""
+
+    def __init__(self, var_name, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.var_name = var_name
+
+    @property
+    def desc(self):
+        return f"variable {self.var_name!r} may be shadowed, {self.lines_str}"
+
+
+class DeclarationShadowedCheck(Check):
+    """Scan ebuilds for shadowed variable assignments in global scope."""
+
+    _source = sources.EbuildParseRepoSource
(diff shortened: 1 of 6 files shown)


src/pkgcheck/checks/codingstyle.py
```diff
@@ -1397,3 +1397,43 @@ def feed(self, pkg):
...
+class VariableShadowed(results.LinesResult, results.Warning):
+    """Variable is shadowed or repeatedly declared. This is a possible typo."""
+
+    def __init__(self, var_name, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.var_name = var_name
+
+    @property
+    def desc(self):
+        return f"possible variable {self.var_name!r} is shadowed, {self.lines_str}"
```
thesamesam, line 1411:
```suggestion
        return f"Variable {self.var_name!r} appears to be shadowed, {self.lines_str}"
```
or
```suggestion
        return f"Variable {self.var_name!r} is shadowed, {self.lines_str}"
```
or
```suggestion
        return f"Variable {self.var_name!r} may be shadowed, {self.lines_str}"
```

src/pkgcheck/checks/codingstyle.py
```diff
@@ -1397,3 +1397,43 @@ def feed(self, pkg):
...
+
+
+class VariableShadowedCheck(Check):
+    """Scan ebuilds shadowed variable assignments in global scope."""
```
thesamesam, line 1415:
```suggestion
    """Scan ebuilds for shadowed variable assignments in global scope."""
```

src/pkgcheck/checks/codingstyle.py
thesamesam: I think I like the option 2 or the last one?

arthurzam: I think the last one is simpler sounding between the two.

thesamesam (approved):

Issue #630, opened 2023-10-01
[New Check]: Detect invalid use of `has_version --host-root` with EAPI 7


thesamesam:
### Is there an existing such new check request for this?

- [X] I have searched the existing issues

### Explain

EAPI 7 bans `has_version --host-root` in favour of using the relevant dependency class (like `-b`).

### Examples

https://bugs.gentoo.org/915019

https://github.com/gentoo/gentoo/blob/ddbbca3c064b767d9dab6b63a6046537f8ebc01e/dev-java/openjfx/openjfx-11.0.11_p1-r1.ebuild#L131 has:
```ebuild
	if has_version --host-root dev-java/openjdk:${SLOT}; then
		export JAVA_HOME=${EPREFIX}/usr/$(get_libdir)/openjdk-${SLOT}
		export JDK_HOME="${JAVA_HOME}"
		export ANT_RESPECT_JAVA_HOME=true
       [...]
   fi
```

### Output message

BadHasVersionArgument

### Documentation

has_version used with illegal argument

### Result level

error

meklu:
> https://github.com/gentoo/gentoo/blob/ddbbca3c064b767d9dab6b63a6046537f8ebc01e/dev-java/openjfx/openjfx-11.0.11_p1-r1.ebuild#L131 has:
> 
> ```shell
> 	if has_version --host-root dev-java/openjdk:${SLOT}; then
> 		export JAVA_HOME=${EPREFIX}/usr/$(get_libdir)/openjdk-${SLOT}
> 		export JDK_HOME="${JAVA_HOME}"
> 		export ANT_RESPECT_JAVA_HOME=true
>        [...]
>    fi
> ```

Just a couple of lines later https://github.com/gentoo/gentoo/blob/ddbbca3c064b767d9dab6b63a6046537f8ebc01e/dev-java/openjfx/openjfx-11.0.11_p1-r1.ebuild#L138 has the following:
```shell
			JDK_HOME=$(best_version --host-root dev-java/openjdk-bin:${SLOT})
			[[ -n ${JDK_HOME} ]] || die "Build VM not found!"
```
Judging by the trace in the bugzilla bug these `{has,best}_version` functions share code, so you should probably be looking for `best_version --host-root` as well. The query function reference being silent about the old option seems to reinforce this notion (<https://devmanual.gentoo.org/function-reference/query-functions/>).

ulm:
[...]

PR #631, opened 2023-10-01
BannedEapiCommand: check for has_version --host-root

arthurzam:
Catch ``has_version --host-root`` and ``best_version --host-root`` calls, which are not allowed in EAPI>=7. Use pkgcore EAPI's option `query_host_root` which is False for EAPI>=7, so great flag for our case.

On current master I only catch:

```
net-misc/zerotier
  BannedEapiCommand: version 1.6.4: 'has_version --host-root' banned in EAPI 7, used on line 38: 'has_version --host-root "sys-devel/clang:${LLVM_SLOT}"'
  BannedEapiCommand: version 1.6.4: 'has_version --host-root' banned in EAPI 7, used on line 43: 'has_version --host-root "=sys-devel/lld-${LLVM_SLOT}*"'
```

Resolves: https://github.com/pkgcore/pkgcheck/issues/630

### DIFF
--- a/src/pkgcheck/checks/codingstyle.py
+++ b/src/pkgcheck/checks/codingstyle.py
@@ -91,6 +91,15 @@ class BadCommandsCheck(Check):
                     )
                 elif name in pkg.eapi.phases.values():
                     yield BannedPhaseCall(line=name, lineno=lineno + 1, pkg=pkg)
+                elif name in ("has_version", "best_version"):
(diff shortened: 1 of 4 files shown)


thesamesam (approved):
Thanks! We should check https://dev.gentoo.org/~ulm/pms/head/pms.html#section-12.3.2 for other missing bits.

Issue #636, opened 2023-11-23
Ability to blacklist urls for MissingRemoteId


parona-source:
I provide go vendor tarballs through gitlab packages so it generates bogus MissingRemoteId for that url.

app-misc/spicetify-cli
  MissingRemoteId: missing <remote-id type="gitlab">api/v4/projects/37881342/packages/generic/spicetify-cli/2.27.1/spicetify-cli-2.27.1-deps.tar.xz</remote-id> (inferred from URI 'https://gitlab.com/api/v4/projects/37881342/packages/generic/spicetify-cli/2.27.1/spicetify-cli-2.27.1-deps.tar.xz')
  
  So the ability to ignore this would be good be good. Or at the very least pkgcheck could be smarter about not seeing gitlab API urls as repos.

sin-ack:
Hit this issue when building vendored Zig packages in gentoo/gentoo#39857.

PR #637, opened 2023-11-26
MissingRemoteId: MissingRemoteId: expand gitlab matching rules

parona-source:
* Notably handles the particular case with gitlab package hosting.

Bug: https://github.com/pkgcore/pkgcheck/issues/636


parona-source:
Apologies for the noise. Didn't realise subprojects were a thing.

parona-source:
Changedapproach to include the reserved words in the regex.

PR #638, opened 2023-11-27
checks.git: Set tarfile filter

parona-source:
* Not setting it triggers a DeprecationWarning in python3.12.
* Despite tarballs by git-archive being trusted, there isn't benefit in trusting it as the tarballs don't use metadata that would require it.


arthurzam (approved):
Thank you

PR #641, opened 2023-12-08
OutdatedProfilePackage: don't warn when version was removed not long ago

arthurzam:


thesamesam (approved):
lgtm but probably worth a test

PR #656, opened 2024-01-23
author/CD fix, fixes for reliance on `category_dirs` and that being fundamentally broken in pkgcore

ferringb:
Rough summary:
* re-add myself as an author, disable CD doc deploy for forks.
* remove reliance on `category_dirs` from EbuildRepo.  See https://github.com/pkgcore/pkgcore/blob/677649351c171685d332b9d9e500da31baf33a74/src/pkgcore/ebuild/repository.py#L527 ; if no categories were inherited or declared, pkgcore violates PMS via falling back to a listdir of the repo root (see https://github.com/pkgcore/pkgcore/blob/677649351c171685d332b9d9e500da31baf33a74/src/pkgcore/ebuild/repository.py#L506 ).  As said, this violates PMS.  Remove reliance on it in pkgcheck in light of that.
* Add a check to pkgcheck for the unlikely scenario where someone has a file in the place of a directory for categories.  Both PM's silently swallow this scenario, hence the check.

There probably should be an explicit check added for if a standalone repository doesn't declare any categories, but that would be a seperate PR.


arthurzam (approved):
Looks great, thank you.

(sorry if it was me dropping you from authors)

ferringb:
> Looks great, thank you.
> 
> (sorry if it was me dropping you from authors)

Yeah, I figured it was the pyproject.toml transfer.  Shit happens, even if I grumble mildly during it... so no worries ;)

PR #657, opened 2024-01-25
Update git blame ignore list, add mgórny and marienz to authors

ferringb:
Same as https://github.com/pkgcore/pkgcore/pull/429 ; I manually audited the history for the rev list and for double checking on who had what contributions, and what contributions still live on despite code reshuffling and variable renames obscuring it.  Marienz is the one who still has intrinsics underneath the various additions and fingerprints.  Not a lot, but core components/contributions.  Things like use addon or metadata checks, although again- it's obscured by the years.  That's still their copyright however.

Finally, Michaeł Górny should be added for obvious reasons.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,4 +1,32 @@
-# Reformat the repository with black-22.12.0
-ae7dd4184f63185880738c5133f326fe47c6606a
 # Reformat the repository with black-23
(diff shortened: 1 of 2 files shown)


ferringb:
I'd normally commit something like this, but doing PR to ensure no issues from others as to the change.

arthurzam (approved):

PR #658, opened 2024-01-25
docs: Add intersphinx linkage to pkgcore/snakeoil.

ferringb:
For any references that go back into those, they're not reachable for doc rendering without this.


arthurzam (approved):
thank you

PR #661, opened 2024-02-01
addons.net: suppress urllib3 import warnings

CyberTailor:
Most notably, this includes NotOpenSSLWarning:
https://github.com/urllib3/urllib3/issues/3020#issuecomment-1785873825

### DIFF
--- a/src/pkgcheck/addons/net.py
+++ b/src/pkgcheck/addons/net.py
@@ -2,12 +2,14 @@
 
 import logging
 import os
-
-import requests
+import warnings
 
 from ..checks.network import RequestError, SSLError
 
-# suppress all urllib3 log messages
+# suppress all urllib3 log messages and import warnings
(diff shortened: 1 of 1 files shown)


thesamesam:
Why not limit this to the OpenSSL warning, with a comment mentioning LibreSSL?

CyberTailor:
> Why not limit this to the OpenSSL warning, with a comment mentioning LibreSSL?

Because this code:
```python
from urllib3.exceptions import NotOpenSSLWarning
```

throws a warning by itself

arthurzam:
I want some extra background for that. You just want me to hide the warning, I understand, and am fine. But I wanted to test it and failed, including with live urllib3. In my partial understanding, this is triggered when not openssl is used for ssl, but where can this happen? Gentoo supports only openssl from what I know?

Issue #666, opened 2024-02-18
[New Check]: pkgcheck scan: "newer EAPI available"


juippis:
### Is there an existing such new check request for this?

- [X] I have searched the existing issues

### Explain

I wonder how feasible it would be to add a check which tells you when you can bump to latest available EAPI. Apparently we still have devs pushing EAPI-7 ebuilds when 8 would work. And this is seen tons with users who contribute sporadically, not following our latest -dev changes.

### Examples

Someone pushes an EAPI-7 ebuild, while (with no inherits or) the eclasses inherited already supports EAPI-8.

### Output message

Ebuild eligible for newer EAPI: 8

### Result level

warning

PR #670, opened 2024-03-02
NewerEAPIAvailable: committing new ebuilds with old EAPI

arthurzam:
Requested by @juippis, and it indeed seems useful.

Catch cases where new ebuilds are committed with old EAPIs (checks against inherited eclasses). This is checked during `--commits` stage.

Resolves: https://github.com/pkgcore/pkgcheck/issues/666


thesamesam (approved):
Nice and elegant!

PR #684, opened 2024-06-02
Permit the '@' character in filenames

floppym:
This will allow for systemd template units to be stored in files/ without mangling the filename.

PMS says nothing about filenames in files/ so we can really permit whatever we find palatable.

### DIFF
--- a/src/pkgcheck/checks/pkgdir.py
+++ b/src/pkgcheck/checks/pkgdir.py
@@ -17,7 +17,7 @@ allowed_filename_chars = set()
 allowed_filename_chars.update(chr(x) for x in range(ord("a"), ord("z") + 1))
 allowed_filename_chars.update(chr(x) for x in range(ord("A"), ord("Z") + 1))
 allowed_filename_chars.update(chr(x) for x in range(ord("0"), ord("9") + 1))
-allowed_filename_chars.update([".", "-", "_", "+", ":"])
+allowed_filename_chars.update([".", "-", "_", "+", ":", "@"])
 
 
 class MismatchedPN(results.PackageResult, results.Error):


ulm:
The charset in the repo is based on the [POSIX Portable Filename Character Set](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282), with the addition of the plus sign.

IIRC this was discussed in QA at some point because there were files with problematic characters like `&`.

https://devmanual.gentoo.org/general-concepts/tree/index.html#what-belongs-in-the-tree

floppym:
> The charset in the repo is based on the [POSIX Portable Filename Character Set](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282), with the addition of the plus sign.

As I have said before, that character set is overly restrictive. It excludes many characters that work just fine on the platforms Gentoo targets.

> IIRC this was discussed in QA at some point because there were files with problematic characters like &.

The line "Files whose name contains characters outside <c>[A-Za-z0-9._+-]" was added to the devmanual by you in this commit.

https://gitweb.gentoo.org/proj/devmanual.git/commit/?id=57f7cae438731c26130d8f45e12572e727185189

It references the "status quo", not any QA decision.

ulm:
> > IIRC this was discussed in QA at some point because there were files with problematic characters like &.
> 
> The line "Files whose name contains characters outside [A-Za-z0-9._+-]" was added to the devmanual by you in this commit.
> 
> https://gitweb.gentoo.org/proj/devmanual.git/commit/?id=57f7cae438731c26130d8f45e12572e727185189
> 
> It references the "status quo", not any QA decision.

To clarify, not necessarily an official QA decision, but it was discussed. I certainly didn't add that bullet point without getting consensus.

At the time it was also a no-brainer because a) it reflected the status of the tree, and b) the rationale for the POSIX charset was that it was already mentioned [elsewhere in the devmanual](https://devmanual.gentoo.org/ebuild-writing/file-format/index.html#file-naming-rules) (this has been there since 2005, in ciaranm's [RST version](https://gitweb.gentoo.org/proj/devmanual.git/diff/ebuild-writing/file-format/text.rst?h=historical&id=b5eb706b0d006535cb8e48b7a03177249a630dd0)). So, not much thought was given to addition of other characters.

floppym:
> To clarify, not necessarily an official QA decision, but it was discussed.

Yeah, I get triggered by undocumented QA decisions. I'm sure it was discussed by some people, but without a record of it I have no context.

In the interest of keeping things simple/consistent, I added '@' to the existing set of permitted characters and dropped the separate set specific to files/. If that is undesirable, I can revert back.

floppym:
Also for the purposes of documentation, this devmanual PR was merged today. https://github.com/gentoo/devmanual/pull/339

ulm:
Should the git hook allow `@` also for directory names? Currently it doesn't:
https://cgit.gentoo.org/infra/githooks.git/commit/local?id=e24eacaa31707ef669cd5a9747a1bf08a5759fdf

floppym:
I'm giving up on this for now. This QA check is coded in too many places and I'm just too frustrated at this point.

PR #707, opened 2024-10-11
python: Fixes for dev-lang/python-freethreading support

mgorny:
gentoo/gentoo#38918

### DIFF
--- a/src/pkgcheck/checks/python.py
+++ b/src/pkgcheck/checks/python.py
@@ -651,7 +651,8 @@ class PythonCompatCheck(Check):
                 (
                     f"python{x.slot.replace('.', '_')}"
                     for x in deps
-                    if x.key == "dev-lang/python" and x.slot is not None
+                    if x.key in ("dev-lang/python", "dev-lang/python-freethreading")
+                    and x.slot is not None
                 ),
                 key=self.sorter,
             )[-1]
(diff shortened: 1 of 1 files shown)


src/pkgcheck/checks/python.py
```diff
@@ -52,7 +52,7 @@ def is_python_interpreter(pkg):
         # ignore python:2.7 deps since they are being phased out from eclass
         # support
         return pkg.slot is None or not pkg.slot.startswith("2")
-    return pkg.key in ("dev-python/pypy3",)
+    return pkg.key in ("dev-python/pypy3", "dev-lang/python-freethreading")
```
thesamesam, line 55: Don't we need to update this to pypy3_10 and so on?

mgorny: No, we depend on `dev-python/pypy3:=`.

thesamesam: Thanks. A comment may help but not required.

mgorny:
Let's put this on hold per https://archives.gentoo.org/gentoo-dev/d3fb6b685e511bce4183b674dd171c9e710765d8.camel@gentoo.org/T/#u. Perhaps we'll go straight for `dev-lang/python3_13t` instead.

arthurzam (approved):
While I understand this is a draft, I see no issue with the way it is done here. I'm also fine with not touching tests for that :)

mgorny:
Given all the opposition on the mailing list, I retract this. Let's shove freethreading into a slot of `dev-lang/python`. Fewer problems for me, can merge it right away and slot lovers get what they want.

PR #724, opened 2025-01-15
OutdatedBlocker: update from 2 years -> 4 years

thesamesam:
2 years is too aggressive IMHO. See de65ca7b4a3f3a68b65b56e2dcaf6e511d3a3443 for a recent example of this in ::gentoo but there's been plenty of such cases in #gentoo on IRC where we got users upgrading fine but they became unstuck by a removed blocker.

Such blockers are really cheap and the purpose of the warning is really to remove ancient cruft rather than actually obstruct people upgrading systems where it's viable.

Note that it's especially problematic because of the current behaviour of python-exec where emerge won't be able to resume if python-exec got updated but portage hasn't yet been rebuilt w/ new PYTHON_TARGETS.

(... because if any interruption occurs, it can be fatal.)

Bug: https://bugs.gentoo.org/916498

### DIFF
--- a/src/pkgcheck/checks/metadata.py
+++ b/src/pkgcheck/checks/metadata.py
@@ -995,7 +995,7 @@ class DependencyCheck(Check):
 
 
 class OutdatedBlocker(results.VersionResult, results.Info):
-    """Blocker dependency removed at least two years ago from the tree.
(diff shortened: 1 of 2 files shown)


arthurzam (approved):

arthurzam:
Missing sign-off :(

PR #729, opened 2025-02-04
Feat: Add suggestion of corrective action on error

martin-kokos:
When error like this happens, try clearing cache
```
pkgcheck scan: error: failed running git log: fatal: Invalid revision range e8f4f3e2bc69cb25f83e07e4e00ff0fee13d2d35..origin/HEAD
```

### DIFF
--- a/src/pkgcheck/addons/git.py
+++ b/src/pkgcheck/addons/git.py
@@ -126,7 +126,10 @@ class GitLog:
         if not self._running:
             if self.proc.poll() or not line:
                 error = self.proc.stderr.read().decode().strip()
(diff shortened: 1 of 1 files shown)


thesamesam:
The commit message should be more precise and detailed. I assumed based on the PR title that this was some grand rework to deliver suggestions in general for `pkgcheck scan` results.

Something like:
```
git: suggest clearing cache on `git log` error

In some cases, the cache gets corrupted, showing:
...

Recommend clearing the cache w/ `pkgcheck cache -R` as a workaround.
```

PR #743, opened 2025-07-09
DependencyCheck: also check `<=` for missing package revision

mgorny:
Extend missing package revision in dependency check to `<=` operator. It is similar to `=` in that it only matches specific revision and not revisions higher than that, and therefore tends to be used mistakenly when people want the version that's not newer than specified version.


arthurzam (approved):
I'll fix it on merge

PR #744, opened 2025-07-11
Replace p7zip with 7zip for .7z files

GoGoOtaku:
P7zip was a fork of 7zip from a time when 7zip was windows only. Since then 7zip got official Linux support and work on p7zip stopped. At this point p7zip is old and outdated and may or may not be vulnerable. (See notice on the official Gentoo wiki)

Last year I filed a bug about this on the Gentoo bug tracker which is currently being worked on.
I think it is somewhat counter productive to tell people to require p7zip when there is active work to replace it.

https://wiki.gentoo.org/index.php?title=P7zip&oldid=1306559
https://bugs.gentoo.org/942397

### DIFF
--- a/src/pkgcheck/checks/metadata.py
+++ b/src/pkgcheck/checks/metadata.py
@@ -1723,7 +1723,7 @@ class MissingUnpackerDepCheck(Check):
     non_system_unpackers = ImmutableDict(
         {
             ".zip": frozenset(["app-arch/unzip"]),
-            ".7z": frozenset(["app-arch/p7zip"]),
+            ".7z": frozenset(["app-arch/7zip"]),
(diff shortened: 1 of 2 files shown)


thesamesam:
Please make the commit message match the PR description so that `git log` is useful, and the information survives if we move from GH.

GoGoOtaku:
Done

I also squashed the commit down

arthurzam:
Missing sign-off (same rules as with Gentoo)

GoGoOtaku:
I never submitted directly to Gentoo.
I figure you meant this rule: https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin
If so I hope this is adequate enough. I'm sorry if it is not.

Issue #745, opened 2025-07-12
[New Check]: Packages from master repository clobbered


mgorny:
### Is there an existing such new check request for this?

- [x] I have searched the existing issues

### Explain

In `::guru`, we don't want packages that are in `::gentoo` already. Back in the `::sunrise` times, we used a custom script for that. Would be nice to have a dedicated check in `pkgcheck` instead — perhaps toggled via `metadata/qa-policy.conf`.

I think it should be "error" since it can clobber packages from `::gentoo`.

### Output message

Package from master repository clobbered, please remove the duplicate

### Documentation

The particular repository is meant to provide additional packages not found in the master repositories. Once a package is added to the master repository, its duplicate should be removed.

### Result level

error

PR #746, opened 2025-07-12
Add a check for clobbered packages in overlays

mgorny:
Add a check that reports if a given package clobbers a package from the master repository.  Since this only applies to some repositores (e.g. ::guru), it is optional (disabled by default).

Closes: https://github.com/pkgcore/pkgcheck/issues/745

### DIFF
--- a/src/pkgcheck/checks/overlays.py
+++ b/src/pkgcheck/checks/overlays.py
@@ -2,7 +2,7 @@ from snakeoil.sequences import iflatten_instance
 from snakeoil.strings import pluralism
 
 from .. import results, sources
-from . import MirrorsCheck, OptionalCheck, OverlayRepoCheck, RepoCheck
+from . import Check, MirrorsCheck, OptionalCheck, OverlayRepoCheck, RepoCheck
 
 
 class UnusedInMastersLicenses(results.VersionResult, results.Warning):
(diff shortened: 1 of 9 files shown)


mgorny:
@arthurzam, any idea how we could make it configured inside repository? I suppose that for the time being we can just explicitly add it to ::guru's CI.

src/pkgcheck/checks/overlays.py
```diff
@@ -132,3 +132,34 @@ def feed(self, pkg):
...
+    @property
+    def desc(self):
+        return f"package from repository ::{self.repository} clobbered"
+
+
+class MasterPackageClobberedCheck(OverlayRepoCheck, OptionalCheck):
+    """Detect clobbering packages from master."""
+
+    _source = sources.PackageRepoSource
+    known_results = frozenset([MasterPackageClobbered])
```
arthurzam, line 158:
```suggestion
    known_results = frozenset({MasterPackageClobbered})
```

it makes it faster to load by something like 100ns from O2 bytecode

src/pkgcheck/checks/overlays.py
```diff
@@ -132,3 +132,34 @@ def feed(self, pkg):
...
+
+    def feed(self, pkgset):
+        for pkg in pkgset:
+            for repo in self.options.target_repo.masters:
+                if repo.has_match(pkg.unversioned_atom):
+                    yield MasterPackageClobbered(str(repo), pkg=pkg)
+            break
```
arthurzam, line 165:
```suggestion
        pkg = pkgset[0]
        for repo in self.options.target_repo.masters:
            if repo.has_match(pkg.unversioned_atom):
                yield MasterPackageClobbered(str(repo), pkg=pkg)
```

it won't enter here with empty `pkgset`

src/pkgcheck/checks/overlays.py
```diff
@@ -132,3 +132,34 @@ def feed(self, pkg):
...
+    def __init__(self, repository: str, **kwarg):
+        super().__init__(**kwarg)
+        self.repository = repository
+
+    @property
+    def desc(self):
+        return f"package from repository ::{self.repository} clobbered"
+
+
+class MasterPackageClobberedCheck(OverlayRepoCheck, OptionalCheck):
```
arthurzam, line 154: If you read `qa-policy.conf` and decide by it, you can drop `OptionalCheck`

arthurzam (changes requested):
> @arthurzam, any idea how we could make it configured inside repository? I suppose that for the time being we can just explicitly add it to ::guru's CI.

Look what we used for `AcctCheck::load_ids_from_configuration`. If current repo doesn't have it enabled, just raise `SkipCheck`

mgorny:
> Look what we used for `AcctCheck::load_ids_from_configuration`. If current repo doesn't have it enabled, just raise `SkipCheck`

Yeah, but I don't want ad-hoc design for now. Perhaps it would be better to have some kind of `optional_pkgcheck_checks` there. Anyway, I'm too tired to pursue this further.

mgorny:
Thinking about it further:

1. We should skip packages if all versions in the master repository are masked (i.e. the package is being moved from `::gentoo` to `::guru`). Not sure how to test that, @arthurzam — perhaps a `fix.patch` that adds it to `package.mask`?
2. I think this should be controlled via `layout.conf` — perhaps as `new-packages-only = true`. We could then also block them from clobbering master from package manager level. I'll mail gentoo-dev later.

arthurzam:
> Thinking about it further:
> 
>     1. We should skip packages if all versions in the master repository are masked (i.e. the package is being moved from `::gentoo` to `::guru`). Not sure how to test that, @arthurzam — perhaps a `fix.patch` that adds it to `package.mask`?
> 

You can just add another package to ::master which is package.masked, and have it at ::guru, and nothing in expected.json. If you remove your code which checks masks - you should see result - and fail (since unexpected result), and with your code, everything passes.

mgorny:
Oh nice. Because of previous failures I've been hitting, I thought `PN` must match an actual keyword — but apparently not.

mgorny:
Updated.

Issue #740, opened 2025-06-23
[New Check]: `EPYTEST_TIMEOUT` overriding user value


mgorny:
### Is there an existing such new check request for this?

- [x] I have searched the existing issues

### Explain

We could check for `EPYTEST_TIMEOUT` set unconditionally, and therefore preventing the user's `make.conf` override from taking effect.

### Examples

https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-python/trimesh/trimesh-4.6.12.ebuild?id=bbfcebd7f38daddea8a68d685ea7ba04db6f1f6c#n52
https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-python/virtualenv/virtualenv-20.31.2.ebuild?id=bbfcebd7f38daddea8a68d685ea7ba04db6f1f6c#n118

### Output message

`EPYTEST_TIMEOUT` does not respect user override

### Documentation

`EPYTEST_TIMEOUT` should be set using the `: "${EPYTEST_TIMEOUT:=...}"` syntax, in order to permit the user to override the timeout via `make.conf` or environment variables.

### Result level

None

Issue #739, opened 2025-06-23
[New Check]: `EPYTEST_*` variables declared after calling `distutils_enable_tests`


mgorny:
### Is there an existing such new check request for this?

- [x] I have searched the existing issues

### Explain

It would be worthwhile to check that people are declaring `EPYTEST_*` variables prior to calling `distutils_enable_tests`. Since the variables are used both by that function and by `epytest` directly, declaring them past `distutils_enable_tests` will still give the "more obvious" result (such as using `pytest-xdist`) but without actually adding the dependencies.

That said, there are a few things to note:

1. It is valid to declare/override `EPYTEST_*` in phase functions, in particular if the dependencies are conditional. I think it would be good enough to catch uses without `local`.
2. We recommend using the `: "${EPYTEST_TIMEOUT:=...}"` form to let the user override the timeout. It would be nice to detect that form as well.
3. Strictly speaking, not all `EPYTEST_*` variables actually need to be set early. However, I think it's fine to catch all of them.

### Output message

`EPYTEST_...` variables declared after calling `distutils_enable_tests`

### Documentation

The `EPYTEST_*` variables need to be declared prior to calling `distutils_enable_tests`, as that function uses them to add dependencies on appropriate pytest plugins. If adding the dependencies is undesirable, it is recommended to declare the variables `local` within the `python_test()` function, to make it clear.

### Result level

warning

PR #747, opened 2025-07-12
python: `EPYTEST_*` related checks

mgorny:
1. Detect `EPYTEST_TIMEOUT` overriding user value (#740)
2. Add a check for misplaced `EPYTEST_*` variables (#739)

### DIFF
--- a/src/pkgcheck/checks/python.py
+++ b/src/pkgcheck/checks/python.py
@@ -266,6 +266,21 @@ class MisplacedEPyTestVar(results.LineResult, results.Error):
         )
 
 
+class ShadowedEPyTestTimeout(results.LineResult, results.Warning):
+    """``EPYTEST_TIMEOUT`` shadows user-specified value
+
+    ``EPYTEST_TIMEOUT`` should be set via ``${EPYTEST_TIMEOUT:=...}`` to permit
+    using an environment variable to override it.
+    """
+
+    @property
+    def desc(self):
+        return (
+            f"line {self.lineno}: EPYTEST_TIMEOUT shadows user value, use "
+            f"${{EPYTEST_TIMEOUT:=...}} instead: {self.line!r}"
+        )
+
+
 class PythonCheck(Check):
     """Python eclass checks.
 
@@ -288,6 +303,7 @@ class PythonCheck(Check):
             PythonAnyMismatchedDepHasVersionCheck,
             PythonMissingSCMDependency,
             MisplacedEPyTestVar,
+            ShadowedEPyTestTimeout,
         }
     )
 
@@ -455,6 +471,13 @@ class PythonCheck(Check):
                 if lineno > det_lineno:
                     line = pkg.node_str(var_node)
                     yield MisplacedEPyTestVar(var_name, line=line, lineno=lineno + 1, pkg=pkg)
+
+        for var_node in bash.var_assign_query.captures(pkg.tree.root_node).get("assign", ()):
+            var_name = pkg.node_str(var_node.child_by_field_name("name"))
+            if var_name == "EPYTEST_TIMEOUT":
+                lineno, _ = var_node.start_point
+                line = pkg.node_str(var_node)
+                yield ShadowedEPyTestTimeout(line=line, lineno=lineno + 1, pkg=pkg)
 
     @staticmethod
     def _prepare_deps(deps: str):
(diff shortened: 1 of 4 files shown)


src/pkgcheck/checks/python.py
```diff
@@ -247,6 +247,40 @@ class PythonMissingSCMDependency(results.VersionResult, results.Warning):
     )
 
 
+class MisplacedEPyTestVar(results.LineResult, results.Error):
+    """Invalid placement of EPYTEST_* variable
```
arthurzam, line 251:
```suggestion
    """Invalid placement of ``EPYTEST_*`` variable
```

src/pkgcheck/checks/python.py
```diff
@@ -247,6 +247,40 @@ class PythonMissingSCMDependency(results.VersionResult, results.Warning):
...
+
+    @property
+    def desc(self):
+        return (
+            f"variable set after calling distutils_enable_tests, line {self.lineno}: {self.line!r}"
+        )
+
+
+class OverriddenEPyTestTimeout(results.LineResult, results.Warning):
+    """EPYTEST_TIMEOUT overrides user-specified value
```
arthurzam, line 270:
```suggestion
    """``EPYTEST_TIMEOUT`` overrides user-specified value
```

src/pkgcheck/checks/python.py
```diff
@@ -247,6 +247,40 @@ class PythonMissingSCMDependency(results.VersionResult, results.Warning):
...
+    ``EPYTEST_TIMEOUT`` should be set via ``${EPYTEST_TIMEOUT:=...}`` to permit
+    using an environment variable to override it.
+    """
+
+    @property
+    def desc(self):
+        return (
+            "EPYTEST_TIMEOUT overrides user value, use ${EPYTEST_TIMEOUT:=...} "
+            f"instead, line {self.lineno}: {self.line!r}"
+        )
```
arthurzam, line 281:
```suggestion
        return (
            f"line {self.lineno}: EPYTEST_TIMEOUT overrides user value, use "
            f"${{EPYTEST_TIMEOUT:=...}} instead: {self.line!r}"
        )
```

src/pkgcheck/checks/python.py
```diff
@@ -389,6 +425,54 @@ def check_pep517(self, pkg):
...
+            if node.children[0].text != b"${":
+                continue
+            name = node.children[1].text
+            if node.children[2].text not in (b"=", b":="):
+                continue
+            yield (name.decode(), node)
+
+    def check_epytest_vars(self, pkg):
+        """Check for incorrect use of EPYTEST_* variables"""
+        # TODO: do we want to check for multiple det calls? Quite unlikely.
```
arthurzam, line 445: Actually I think we should. People could put it just after `*DEPEND` vars, then have PATCHES, long DESELECT, and then an additional `det` call

arthurzam (changes requested):

src/pkgcheck/checks/python.py
mgorny: Yeah, but I think it belongs in a different check — along with things like `llvm.org_set_globals`, i.e. I'd leave it for later. Perhaps even some eclassdoc marker for this class of "call once in global scope" functions.

mgorny:
Updated.

thesamesam (commented):
(Sorry, I didn't realise these comments were still pending and not submitted.)

src/pkgcheck/checks/python.py
```diff
@@ -268,7 +302,9 @@ class PythonCheck(Check):
             PythonAnyMismatchedUseHasVersionCheck,
             PythonAnyMismatchedDepHasVersionCheck,
             PythonMissingSCMDependency,
-        ]
+            MisplacedEPyTestVar,
```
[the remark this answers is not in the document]
mgorny: Yeah, mixed feelings about that too but it seems more consistent than "Epytest" or "EPytest".

src/pkgcheck/checks/python.py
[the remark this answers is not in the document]
mgorny: But we do — but I want to implement it as a separate, more generic check for functions you're supposed to call only once.

src/pkgcheck/checks/python.py
```diff
@@ -247,6 +247,40 @@ class PythonMissingSCMDependency(results.VersionResult, results.Warning):
...
+        self.variable = variable
+
+    @property
+    def desc(self):
+        return (
+            f"variable set after calling distutils_enable_tests, line {self.lineno}: {self.line!r}"
+        )
+
+
+class OverriddenEPyTestTimeout(results.LineResult, results.Warning):
```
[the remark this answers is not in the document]
mgorny: Done.

src/pkgcheck/checks/python.py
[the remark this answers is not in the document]
thesamesam: OK.

arthurzam (approved):

PR #754, opened 2025-07-26
PythonCheck: suggest opportunities for EPYTEST_PLUGINS

mgorny:


testdata/repos/python/PythonCheck/EPyTestPluginsSuggestion/EPyTestPluginsSuggestion-0.ebuild
```diff
@@ -0,0 +1,13 @@
+EAPI=8
+
+DISTUTILS_USE_PEP517=flit
+PYTHON_COMPAT=( python3_10 )
+
+inherit distutils-r1
+
+DESCRIPTION="Ebuild with redundant disable-autoload"
```
thesamesam, line 8: The DESCRIPTION isn't accurate for the test here, right?

PR #757, opened 2025-09-03
network: Add PyPIAttestationAvailableCheck

mgorny:
Companion to https://github.com/gentoo/gentoo/pull/43549. The code's a bit hacky, suggestions welcome.

Current results: [pypi-provenance.txt](https://github.com/user-attachments/files/22122839/pypi-provenance.txt)


arthurzam (approved):

ferringb:
> Okay, I've added a "positive" test. Unfortunately, I haven't been able to figure out how to make a "negative" test work.

By negative assertion, are you referring to trying to flex these [except handlers](https://github.com/pkgcore/pkgcheck/pull/757/files#diff-ef18bcf21170e43383bd0c619a9b5d17c65f3137a302755ce7a01f50232b6a9bR383-R391)?  If not- what are the failures you're trying to assert, at least at a high level?

This futures/execution setup is new to me, but being attestations are immutable- what's your thoughts on locally caching the results of past positive checks in such a fashion that network request doesn't need to be made again if it's in the cache?

mgorny:
Kinda. A test that the result isn't output when attestation URL gives 404.

PR #762, opened 2025-11-25
eclass: add ShadowedEclassPhase for multiple-eclass defined phases

### LINKED ISSUE #377: Detect "shadowed" pkg_{setup,pretend} phases
This is mostly an issue with pkg_{setup,pretend,postinst} but it affects other phase functions too.

For example, `xdg.eclass` exports a `xdg_pkg_postinst`, but if one has...
```
EAPI=8

inherit optfeature xdg

pkg_postinst() {
	optfeature "Foo support" dev-libs/foo
}
```
... then `xdg_pkg_postinst` is never called, which may be because the developer didn't realise `xdg` even exported a `pkg_postinst`.

This is arguably more of a problem with things like `fortran-2.eclass` where we might often have to define `pkg_setup` to check for OpenMP support, but then forget to call `fortran-2_pkg_setup` (example modified from `sci-libs/lis/lis-1.6.5.ebuild`:
```
EAPI=8

inherit toolchain-funcs fortran-2

pkg_pretend() {
	[[ ${MERGE_TYPE} != binary ]] && use openmp && tc-check-openmp
}

pkg_setup() {[
    # (Dropped fortran-2_pkg_setup from original ebuild for purposes of example)
	[[ ${MERGE_TYPE} != binary ]] && use openmp && tc-check-openmp && FORTRAN_NEED_OPENMP=1
}
```

thesamesam:
Detect the basic case of an ignored eclass phase where two eclasses are
inherited, both defining the same phase, and the ebuild does not define
a custom implementation of that phase at all. This means one of the exported
phases from the eclasses is being ignored.

Ignore some eclasses with a blacklist where they are known to vary their
API by EAPI or eclass variables, at least for now, because the eclass
cache we have accessible here isn't keyed by EAPI or the context of the
sourcing ebuild.

Bug: https://bugs.gentoo.org/516014
Bug: https://bugs.gentoo.org/795006
Closes: https://github.com/pkgcore/pkgcheck/issues/377

### DIFF
--- a/src/pkgcheck/checks/eclass.py
+++ b/src/pkgcheck/checks/eclass.py
@@ -331,41 +331,43 @@ class EclassUsageCheck(Check):
             yield ShadowedEclassPhase(missing, sorted(exported_phases[missing]), pkg=pkg)
 
     def feed(self, pkg):
-        if pkg.inherit:
-            inherited: set[str] = set()
-            inherits: list[tuple[list[str], int]] = []
-            for node in bash.cmd_query.captures(pkg.tree.root_node).get("call", ()):
-                name = pkg.node_str(node.child_by_field_name("name"))
-                if name == "inherit":
-                    call = pkg.node_str(node)
-                    # filter out line continuations and conditional inherits
-                    if eclasses := [x for x in call.split()[1:] if x in pkg.inherit]:
-                        lineno, _colno = node.start_point
-                        if not inherited and eclasses[0] == pkg.inherit[0]:
-                            inherits.append((eclasses, lineno))
+        if not pkg.inherit:
+            return
 
-                        for eclass in eclasses:
-                            if eclass not in inherited:
-                                inherited.add(eclass)
-                            else:
-                                yield DuplicateEclassInherit(
-                                    eclass, line=call, lineno=lineno + 1, pkg=pkg
-                                )
+        inherited: set[str] = set()
+        inherits: list[tuple[list[str], int]] = []
+        for node in bash.cmd_query.captures(pkg.tree.root_node).get("call", ()):
+            name = pkg.node_str(node.child_by_field_name("name"))
+            if name == "inherit":
+                call = pkg.node_str(node)
+                # filter out line continuations and conditional inherits
+                if eclasses := [x for x in call.split()[1:] if x in pkg.inherit]:
+                    lineno, _colno = node.start_point
+                    if not inherited and eclasses[0] == pkg.inherit[0]:
+                        inherits.append((eclasses, lineno))
 
-            yield from self.check_provided_eclasses(pkg, inherits)
-            yield from self.check_user_variables(pkg, inherits)
-            # verify @PRE_INHERIT variable placement
-            yield from self.check_pre_inherits(pkg, inherits)
-            # verify @DEPRECATED variables or functions
-            yield from self.check_deprecated_variables(pkg, inherits)
-            yield from self.check_deprecated_functions(pkg, inherits)
-            yield from self.check_exported_eclass_phase(pkg, inherits)
+                    for eclass in eclasses:
+                        if eclass not in inherited:
+                            inherited.add(eclass)
+                        else:
+                            yield DuplicateEclassInherit(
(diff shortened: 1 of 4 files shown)


src/pkgcheck/checks/eclass.py
```diff
@@ -254,6 +274,40 @@ def check_provided_eclasses(self, pkg, inherits: list[tuple[list[str], int]]):
...
+        defined_phases = []
+        for node in bash.func_query.captures(pkg.tree.root_node).get("func", ()):
+            func_name = pkg.node_str(node.child_by_field_name("name"))
+            if func_name in known_phases:
+                defined_phases.append(func_name)
+
+        # Strip out phases we already define (even if inside of those, we don't
+        # actually call exported phases from all eclasses inherited). Assume that
+        # a custom phase in the ebuild is intentionally omitting them.
+        missing_custom_phases = set(
```
ferringb: you want this:
```
missing_custom_phases = set(phase for phase, eclass in exported_phases.items() if len(eclass) > 1) 
missing_custom_phases.difference_update(defined_phases)
```

That cuts the set creation down to one, rather than the 3 that's implicit in this.  That form's not a hard requirement, my point is more about the generated and discarded objects in regards to efficiency and muscle memory.

thesamesam:
There's a lot of noise from things like `cargo.eclass` where we only use it for setting up flags but we don't want the other phases sometimes, ditto go. Not sure how to handle that (a heuristic would be fine).

One might argue this really means we should have `xyz-utils.eclass` more often rather than making an eclass do many things, sometimes with toggles to customise it.

We can also have this be off-by-default, but nonetheless, better heuristics would be useful..

src/pkgcheck/checks/eclass.py
```diff
@@ -254,41 +277,97 @@ def check_provided_eclasses(self, pkg, inherits: list[tuple[list[str], int]]):
...
+        # XXX: Some eclasses vary their API based on the EAPI, usually to
+        # 'unexport' a phase. self.eclass_cache is generated once per eclass,
+        # not (eclass, EAPI), so it can't handle this. Ditto phases which are only
+        # exported if a variable is set. Blacklist such eclasses here as it
+        # doesn't happen often.
+        #
+        # We could maybe make this more finely-grained for phases we know
+        # are conditionally exported if this list is impacting coverage
+        # severely.
+        blacklisted_eclasses = ["pypi", "vala", "xdg"]
```
ferringb: Suppressions like this need to be a class member so testing code can vary it.  Plus it's easier to spot. :)

That said- I assume we can rip out a simple literal asignment out of the parsed bash?  I've not played w/ tree-sitter (yet), but I'm hoping we can reliable scrape a global assignment like this, out of  an eclass:
```
_PKGCHECK_DISABLED_QA_CHECKS_=( ShadowedEclassPhase OtherCheck1 OtherCheck2 )
```
I'm suggesting an array specifically so other checks can move their suppression lists into the source itself.  

Doing this also allows external eclasses's to get in on the game- I'm assuming somewhere buried in the check pathways there is a `if it's not the gentoo repo, don't run me` ?

ferringb (commented):
random code poking.  Take what makes sense, discard what doesn't.  Y'all know this code better than I do.

ferringb:
@thesamesam I figure it's still relevant- how high priority is it to land this?  Wait for others with time, have one of us sort it, etc?

Per ticket traffic you've likely seen, I'm trying to clean things up and nuke anything stale or 'wontfix'

thesamesam:
I think it's kind of a neat idea, but I feel like the limitations (inability to say when something varies with EAPI; need some eclassdoc tag for that?) mean we should probably leave it for now. Let's call it WONTFIX for now and I can always revive it if I get some inspiration.

(And thanks for doing that cleanup)

ferringb:
The limitations are beyond my current knowledge, but if you think it *could* be doable- and is worth it- move it to a ticket and tag it with whatever tag we have for "needs help" imo.

I'm not looking to boot good ideas, just institute some kanban type management and ability for us to show up for a day or two, every few weeks, and do some work.

PR #771, opened 2026-02-14
network: add DetachedSignatureAvailableCheck

parona-source:
@thesamesam was interested in this

I couldn't fully follow precedent with the network tests. The check deals with multiple network requests in one go so just a mocking of the returnvalue wasn't enough.


thesamesam (approved):
Thanks a lot for this. It works well for me as well.

src/pkgcheck/checks/network.py
```diff
@@ -465,3 +465,124 @@ def schedule(self, pkg, executor, futures):
...
+            if any(
+                (f.filename.endswith(extension) or f"{f.filename}{extension}" in filenames)
+                for extension in self.detached_signature_extensions
+            ):
+                continue
+
+            for url in f.uri:
+                for extension in self.detached_signature_extensions:
+                    yield (f.filename, f"{url}{extension}")
+        return []
```
arthurzam:
```suggestion
```

no need for this return

src/pkgcheck/checks/network.py
```diff
@@ -465,3 +465,124 @@ def schedule(self, pkg, executor, futures):
...
+    _source = sources.LatestVersionRepoSource
+
+    known_results = frozenset(
+        {
+            DetachedSignatureAvailable,
+            SSLCertificateError,
+        }
+    )
+
+    detached_signature_extensions = [".asc", ".minisig", ".sig", ".sign", ".sigstore"]
```
arthurzam:
```suggestion
    detached_signature_extensions = (".asc", ".minisig", ".sig", ".sign", ".sigstore")
```

src/pkgcheck/checks/network.py
```diff
@@ -465,3 +465,124 @@ def schedule(self, pkg, executor, futures):
...
+        exc = future.exception()
+        if exc is not None:
+            # traceback can't be pickled so serialize it
+            tb = traceback.format_exc()
+            # return exceptions that occurred in threads
+            self.results_q.put(tb)
+            return
+
+        result = future.result()
+        if result is not None:
```
arthurzam:
```suggestion
        if (result := future.result()):
```

src/pkgcheck/checks/network.py
```diff
@@ -465,3 +465,124 @@ def schedule(self, pkg, executor, futures):
...
+            # Don't check for detached signatures if any detached signature is already present for the filename.
+            if any(
+                (f.filename.endswith(extension) or f"{f.filename}{extension}" in filenames)
+                for extension in self.detached_signature_extensions
+            ):
+                continue
+
+            for url in f.uri:
+                for extension in self.detached_signature_extensions:
+                    yield (f.filename, f"{url}{extension}")
```
arthurzam:
```suggestion
                    yield f.filename, url + extension
```

arthurzam (changes requested):
Some small changes I want, but looks really good.

I even want to run it global wise to know if it finds something

src/pkgcheck/checks/network.py
parona-source: This is also done this way in PyPIAttestationAvailableCheck. Should that be changed as well?

I mostly copied that check.

src/pkgcheck/checks/network.py
parona-source: _UrlCheck and PypiAttestationAvailableCheck do this as well. Should I touch these?

src/pkgcheck/checks/network.py
parona-source: Applied suggestion. Didn't touch PyPIAttestationAvailableCheck

Edit: Touched it by minimising it as well.

src/pkgcheck/checks/network.py
parona-source: Realised I could minimise duplicated code by using _UrlCheck. Didn't touch them but I removed this from the new check.

PypiAttestationAvailableCheck looks like it could get minimised as well.

Edit: Touched them.

src/pkgcheck/checks/network.py
```diff
@@ -179,16 +179,14 @@ def _ftp_check(self, attr, url, *, pkg):
...
+        if exc := future.exception():
             # traceback can't be pickled so serialize it
             tb = traceback.format_exc()
             # return exceptions that occurred in threads
             self.results_q.put(tb)
             return
 
-        result = future.result()
-        if result is not None:
+        if result := future.result():
```
ferringb, line 189: Is there any scenario where result can validly be `not bool(future.result())` for a returned result?

In terms of actual 'api' safety, a check of `future.done()` is safer, but it's also a more obnoxious chain of calls..

I mention it in reading the code; this is neither a "must" nor a "should"; just an observation since this is being converted away from `is None`

src/pkgcheck/checks/network.py
```diff
@@ -360,7 +357,7 @@ def desc(self):
         )
 
 
-class PyPIAttestationAvailableCheck(NetworkCheck):
+class PyPIAttestationAvailableCheck(_UrlCheck):
```
ferringb, line 360: unrelated to your  change, but we should promote `_UrlCheck` to a public base, albeit after lacing ABC through it.

src/pkgcheck/checks/network.py
```diff
@@ -464,4 +427,96 @@ def schedule(self, pkg, executor, futures):
                     return
 
         for filename, url in self._get_urls(pkg):
-            self._schedule_check(filename, url, executor, futures, pkg=pkg)
+            self._schedule_check(self._provenance_check, filename, url, executor, futures, pkg=pkg)
+
+
+class DetachedSignatureAvailable(results.VersionResult, results.Info):
```
ferringb, line 433: Just FYI- none of this inheritance chain currently sets `__slots__`, but they'll be converted over.  That's both to prevent accidental access, and for serialization speed reasons..

No change needed here, again, just FYI.

src/pkgcheck/checks/network.py
```diff
@@ -464,4 +427,96 @@ def schedule(self, pkg, executor, futures):
...
+        ".sigstore",
+    )
+
+    def __init__(self, *args, use_addon, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fetch_filter = use_addon.get_filter("fetchables")
+
+    def _verifysig_check(self, filename, url, *, pkg):
+        """Check for typical verify sig URLS."""
+        result = None
```
ferringb, line 474: Thoughts on converting this flow to just be returns?  Your code is fine- there's no possible uninitialized scenario, but if you just lace returns in, the possibility goes away in full while simplifying the control flow considerations.

IE.
```
try:
  # Need redirects to deal with the variance of file servers and urls
  response = self.session.head(url, allow_redirects=True)
except RequestError:
  # should this be debug logged, or does something further down the chain log  it?
  return None
except SSLError as e:
  return SSLCertificateError("SRC_URI", url, str(e), pkg=pkg)
  
content_type = response.headers.get("Content-Type")
# Filtering out text/html matches is useful due to possible false matches with authentication
if (
  response.ok
  and content_type is not None
  and not content_type.startswith("text/html")
):
  return DetachedSignatureAvailable(filename, url, pkg=pkg)
return None
```

That's just an observation of a possible improvement.

ferringb:
The code looks fine- these are just potential points of polish or questions about being brutally paranoid about robustness ;)

> I couldn't fully follow precedent with the network tests. The check deals with multiple network requests in one go so just a mocking of the returnvalue wasn't enough.

I don't know that code in question, but that sounds like something our test utilities should be enhanced to support.  Mind cutting a ticket laying out what you think it should be?  IE, what would've allowed you to use it rather than having to sidestep it?

src/pkgcheck/checks/network.py
ferringb: Actually, one tweak here simplifies things:
```
# Filtering out text/html matches is useful due to possible false matches with authentication
if (
  response.ok
  and not response.headers.get("Content-Type", "").startswith("text/html")
):
```
You can skip the intermediate `content_type`

parona-source:
> > I couldn't fully follow precedent with the network tests. The check deals with multiple network requests in one go so just a mocking of the returnvalue wasn't enough.
> 
> I don't know that code in question, but that sounds like something our test utilities should be enhanced to support. Mind cutting a ticket laying out what you think it should be? IE, what would've allowed you to use it rather than having to sidestep it?

I meant that with other network tests are simpler with having only one request triggered. So instead of simply setting requests = ... etc in responses.py,  I had  to have a function where I kept the implementation details for Session in mind. (Which is where `@contextmanager` and yield come from).

Example:
https://github.com/pkgcore/pkgcheck/blob/master/testdata/repos/network/MetadataUrlCheck/HttpsUrlAvailable/responses.py

Session:
https://github.com/pkgcore/pkgcheck/blob/master/src/pkgcheck/addons/net.py#L36

I don't see an issue with it as is, it currently means there is some ready made boilerplate for the next time its needed.

PR #775, opened 2026-03-22
utils: replace chardet with charset-normalizer

CyberTailor:
chardet 7.0 has been [completely rewritten using AI coding agents](https://github.com/chardet/chardet/issues/327), and now related pkgcheck's tests fail. It's unlikely that chardet v7 will be packaged in Gentoo, so it makes no sense to make pkgcheck compatible with it. It also doesn't really make sense to pin to `chardet<7`, as it won't receive updates. Switching to another library would be fine.

I think we can also remove most of our custom "binary or not" heuristics, [`charset_normalizer.is_binary`](https://charset-normalizer.readthedocs.io/en/stable/api.html#charset_normalizer.is_binary) should be good enough.

Another reason to switch from chardet:
```
$ time python -c "import chardet"

real    0m0,388s
user    0m0,236s
sys     0m0,053s

$ time python -c "import charset_normalizer"

real    0m0,119s
user    0m0,091s
sys     0m0,018s
```

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -327,7 +327,7 @@ epub_exclude_files = ["search.html"]
 
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3", None),
-    "chardet": ("https://chardet.readthedocs.io/en/latest", None),
+    "charset-normalizer": ("https://charset-normalizer.readthedocs.io/en/stable/", None),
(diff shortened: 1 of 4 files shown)


arthurzam (approved):
I see that `charset-normalizer` has no python lib dependencies, which is nice and what I prefer.

Looks good, thank you.

PR #776, opened 2026-03-28
PythonPackageNameCheck: fix TypeError on empty PyPi remote

falbrechtskirchinger:
Skip check when the PyPI `remote-id` in `metadata.xml` is empty and add a test package.

Addresses the following crash observed with a template `metadata.xml` (`<remote-id type="pypi"></remote-id>`):

```
$ pkgcheck scan .
pkgcheck scan: error: Traceback (most recent call last):
  File "/usr/lib/python3.14/site-packages/pkgcheck/pipeline.py", line 191, in _run_checks
    if results := sorted(
                  ~~~~~~^
        chain.from_iterable(pipes[i][-1][scope][j].run(restrict) for j in runners)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ):
    ^
  File "/usr/lib/python3.14/site-packages/pkgcheck/runners.py", line 71, in run
    yield from check.feed(item)
  File "/usr/lib/python3.14/site-packages/pkgcheck/checks/python.py", line 1114, in feed
    if pkg.package != normalize(pypi_name):
                      ~~~~~~~~~^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/pkgcheck/checks/python.py", line 1111, in normalize
    return PROJECT_SYMBOL_NORMALIZE_RE.sub("-", project).lower()
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
TypeError: expected string or bytes-like object, got 'NoneType'
```

Expected result:
```
PkgMetadataXmlEmptyElement: metadata.xml: empty element 'remote-id' on line 9
```

The test is primarily a regression test, for which I could find no precedent. `python/dev-python/PythonMismatchedPackageName4` seems a bit out of place. Should I drop it, put it elsewhere, ...?

### DIFF
--- a/src/pkgcheck/checks/python.py
+++ b/src/pkgcheck/checks/python.py
@@ -1102,6 +1102,12 @@ class PythonPackageNameCheck(Check):
         if len(pypi_remotes) != 1:
             return
 
+        pypi_name = pypi_remotes[0].name
+
+        # skip empty remotes
+        if pypi_name is None:
+            return
+
         def normalize(project: str) -> str:
             """
             Normalize project name using PEP 503 rules
@@ -1110,6 +1116,5 @@ class PythonPackageNameCheck(Check):
             """
             return PROJECT_SYMBOL_NORMALIZE_RE.sub("-", project).lower()
 
-        pypi_name = pypi_remotes[0].name
         if pkg.package != normalize(pypi_name):
             yield PythonMismatchedPackageName(normalize(pypi_name), pkg=pkg)
--- a/testdata/repos/python/dev-python/PythonMismatchedPackageName4/PythonMismatchedPackageName4-0.ebuild
+++ b/testdata/repos/python/dev-python/PythonMismatchedPackageName4/PythonMismatchedPackageName4-0.ebuild
@@ -1,0 +1,9 @@
+# Copyright 2026 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+DESCRIPTION="Ebuild with empty PyPi remote-id in metadata.xml"
+HOMEPAGE="https://github.com/pkgcore/pkgcheck"
+LICENSE="BSD"
+SLOT="0"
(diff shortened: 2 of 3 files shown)


src/pkgcheck/checks/python.py
```diff
@@ -1102,6 +1102,12 @@ def feed(self, pkgs):
         if len(pypi_remotes) != 1:
```
ferringb, line 1102: Unrelated- this should be a warning IMO so we can isolate how to handle this better.  That situation shouldn't exist- if it does, it points at a bad metadata.xml definition, or something (again) we should address.

ferringb:
I think you're doing the inverse of what is needed- isn't this a malformed metadata.xml per [glep68](https://www.gentoo.org/glep/glep-0068.html?rdfrom=https%3A%2F%2Fwiki.gentoo.org%2Findex.php%3Ftitle%3DGLEP%3A68%26redirect%3Dno#id21)?  the 'value'- name- is required.  If they've specified a remote-id but left that out, it's malformed.

falbrechtskirchinger:
> I think you're doing the inverse of what is needed- isn't this a malformed metadata.xml per [glep68](https://www.gentoo.org/glep/glep-0068.html?rdfrom=https%3A%2F%2Fwiki.gentoo.org%2Findex.php%3Ftitle%3DGLEP%3A68%26redirect%3Dno#id21)? the 'value'- name- is required. If they've specified a remote-id but left that out, it's malformed.

While I see your point regarding GLEP 68, my goal with this PR is just to stop `pkgcheck` from crashing outright.

The crash prevents `PackageMetadataXmlCheck` from catching the empty element. While `PkgMetadataXmlEmptyElement` may not fail CI, that is a separate problem that another PR can address.

falbrechtskirchinger:
Also:

https://github.com/pkgcore/pkgcheck/blob/9ecf675fe28137bf16b30fd433b02d56113d4a56/src/pkgcheck/checks/metadata_xml.py#L572-L576

ferringb:
> While I see your point regarding GLEP 68, my goal with this PR is just to stop `pkgcheck` from crashing outright.
> 
> The crash prevents `PackageMetadataXmlCheck` from catching the empty element. While `PkgMetadataXmlEmptyElement` may not fail CI, that is a separate problem that another PR can address.

No argument that this shouldn't fail out, so your PR is merged.

Offhand- this shouldn't even be *necessary*, but pkgcore is letting through schema violating objects- see pkgcore/pkgcore#466 .  I added a commit tweaking your comment to clarify this is suppressing invalid xml, and linking to the upstream.

Thanks!

PR #786, opened 2026-07-05
checks/profiles: sort USE and USE_EXPAND flags in output

laumann:
For more reliable output (especially in gentoo CI) ensure that flags are always sorted in checks that report unknown USE or USE_EXPAND values. In many cases, the values were already sorted, this just moves the sorting into a central location to capture some missed spots.


arthurzam (approved):
thanks

